PR #27, opened 2020-05-04
Gh pages restyle

ewanmay:


src/App.css
```diff
@@ -22,12 +24,32 @@
...
+    align-content: center;
+}
+
+.App-container {
+    height: 100vh;
+    display: flex;
+}
+
+.App-title {
+    font-size: 26px;
```
atatmaja, line 53: Minor: maybe we should start using em for font sizing so it scales better with different screen sizes.

ewanmay: that's a good call

src/components/static/Home.tsx
```diff
@@ -1,7 +1,46 @@
...
+import React, { useState } from "react";
 import './static.css';
+import { Tab, Icon } from "semantic-ui-react";
 
 export default function Home() {
+    const [tabHidden, setTabHidden] = useState(true)
+
+    const panes = [{
+        menuItem: 'Detailed Serosurvey Data (All)',
+        render: () => <Tab.Pane className="col-12 p-0">
```
atatmaja, line 10: Maybe consider abstracting this into a method that accepts 'src' as a parameter since similar logic is used 3 times

PR #35, opened 2020-05-08
Map integration + Bar charts + References

ewanmay:


src/components/Component.js
```diff
@@ -16,14 +16,6 @@ export default function Component() {
         })
     }
 
-    const getAirtableRecords = async () => {
```
atatmaja, line 19: either we can leave this in here for testing or just get rid of Component.js entirely

src/components/charts/Charts.css
```diff
@@ -0,0 +1,32 @@
...
+  font-size: 18px;
+  border: none;
+  background-color: lightgrey; 
+}
+
+.charts-title {
+  font-size: 18px;
+}
+
+.custom-tooltip {
```
atatmaja, line 30: no style definition?

src/components/charts/Charts.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { getAggregateData } from "../../metaAnalysis";
+
+export default function Charts() {
+  const [yAxisSelection, setYAxis] = useState('Population');
+  const [state, dispatch] = useContext(AppContext);
+
+  const { filtered_records } = state;
+  
+  const chartData = getAggregateData(filtered_records, 'country')
+  interface chartData {
```
atatmaja, line 16: We don't seem to be using this interface.

src/components/Component.js
ewanmay: I personally think that it's (for now) fine for testing purposes

src/components/charts/Charts.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  const yAxisOptions = [
+    { key: 'Population', text: 'Population', value: 'Population' },
+    { key: 'Geographies', text: 'Geographies', value: 'Geographies' }
+  ]
+
+  const handleChange = (e: any) => {
+    const { value } = e;
+    setYAxis(value);
```
atatmaja, line 28: why not just setYAxis(e.value)?

src/components/charts/ReferencesTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export default function ReferencesTable() {
+  const [column, setColumn] = useState('');
+  const [state, dispatch] = useContext(AppContext);
+  const [activePage, setActivePage] = useState(1);
+  const [boundaryRange, setBoundaryRange] = useState(1);
+  const [siblingRange, setSiblingRange] = useState(1);
+  const [totalPages, setTotalPages] = useState(10);
+
+    const handlePaginationChange = (e : any, event: any) => {
```
atatmaja, line 15: maybe document why there need to be two events in the function handle

ewanmay: I don't know why, but for some reason this is how it comes

src/components/charts/Charts.css
ewanmay: didn't put any work into styling it but as soon as we do all we need to do is fill that class, it's already connected

src/components/charts/ReferencesTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const handlePaginationChange = (e : any, event: any) => { 
+      const { activePage } = event;
+      setActivePage(activePage)
+    }
+
+
+  const [direction, setDirection] = useState('ascending');
+  const [data, setData] = useState(state.filtered_records);
+
+  const handleSort = (clickedColumn: string) => () => {
```
atatmaja, line 24: why () => () => {}??

src/components/charts/ReferencesTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (column !== clickedColumn) {
+      setColumn(clickedColumn);
+      setDirection('ascending');
+      return
```
atatmaja, line 28: This works, but maybe more readable as an if/else

ewanmay: Yeah true, this isn't returning a value so that makes sense for if else

src/httpClient.ts
```diff
@@ -19,7 +19,7 @@ export default class httpClient {
     }
 
     async getAirtableRecords() {
-        const response = await this.httpGet('http://localhost:5000/airtable_scraper/records');
+        const response = await this.httpGet('/airtable_scraper/records');
```
atatmaja, line 22: Let's set the url prefix (i.e. `http://localhost:5000`) as an environment variable and run a substitution in the generic httpGet method.

src/components/charts/ReferencesTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    setDirection(direction === 'ascending' ? 'descending' : 'ascending');
+  }
+
+  useEffect(() => {
+    const newData = _.sortBy(state.filtered_records, [column]);
```
atatmaja, line 35: maybe we should push all useEffect functions to the top of the component (right below useState/useContext) for better visibility

src/metaAnalysis.ts
```diff
@@ -17,7 +17,7 @@ function calcPooledPrevalence(records: AirtableRecord[]) {
     });
 
     return {
-        seroprevalence: p_over_var_sum / inv_var_sum,
+        seroprevalence: p_over_var_sum / inv_var_sum * 100,
         error: Z_SCORE * Math.sqrt(var_sum)
```
atatmaja, line 21: Also need to multiply error by 100

atatmaja:
Also pls check failing tests/CI build

src/httpClient.ts
ewanmay: I circumvented the cors issue by creating a proxy in the package.json

atatmaja: ahh yeah I see that now. maybe next thing to look at is parameterizing that somehow (with env var or something) to ease deployment

PR #36, opened 2020-05-08
Updates to meta analysis functions

atatmaja:
Add compute overall number of tests aggregated in meta analysis functions, expose function to calculate pooled metrics for all supplied records.

#34

### DIFF
--- a/src/metaAnalysis.ts
+++ b/src/metaAnalysis.ts
@@ -2,10 +2,11 @@ import { AirtableRecord } from "./types"
 
 const Z_SCORE: number = 1.96;
 
-function calcPooledPrevalence(records: AirtableRecord[]) {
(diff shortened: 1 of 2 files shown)


src/metaAnalysis.ts
```diff
@@ -2,28 +2,31 @@ import { AirtableRecord } from "./types"
 
 const Z_SCORE: number = 1.96;
 
-function calcPooledPrevalence(records: AirtableRecord[]) {
+export function aggregateRecords(records: AirtableRecord[]) {
```
atatmaja, line 5: @ewanmay can use this function to aggregate all records, just pass in all the records you want to aggregate (will probably be state.filtered_records in your case)

PR #37, opened 2020-05-09
set up sidebars initially

ewanmay:


src/App.css
```diff
@@ -10,7 +10,6 @@
     justify-content: space-between;
     font-size: 1.25;
     color: white;
-    max-height: 6vh;
```
atatmaja, line 13: point of this?

ewanmay: it was lookin real weird

ewanmay: if we're setting the height of the page we can leave this relative

PR #38, opened 2020-05-10
Working sidebar

ewanmay:


src/components/sidebar/left-sidebar/CountryList.tsx
```diff
@@ -1,9 +1,32 @@
...
+import React, { useContext } from "react";
+import { AppContext } from "../../../context";
+import { getAggregateData } from "../../../metaAnalysis";
+import './CountryList.css';
+import _ from "lodash";
 
 export default function CountryList() {
+  const [state, dispatch] = useContext(AppContext);
+
+  const aggregatedData = _.sortBy(getAggregateData(state.filtered_records, 'country'), ['seroprevalence']).reverse();
```
atatmaja, line 10: Not sure if we'd rather sort countries by alphabetical order or seroprevalence as you've done here. Maybe a question to bring up during a design review

src/components/sidebar/left-sidebar/TotalStats.tsx
```diff
@@ -1,9 +1,31 @@
...
+import { AppContext } from "../../../context";
+import { aggregateRecords, getAggregateData } from "../../../metaAnalysis";
+import { Statistic } from "semantic-ui-react";
+import './TotalStats.css'
 
 export default function TotalStats() {
+  const [state, dispatch] = useContext(AppContext);
+
+  const { seroprevalence, n } = aggregateRecords(state.filtered_records);
+  const aggregatedData = getAggregateData(state.filtered_records, 'country');
```
atatmaja, line 11: This is an unnecessary computation, there's probably a less wasteful way to compute the number of countries.

Maybe we can leverage some of the filters work to do this better (where we will likely store all country names in our context). If we decide to go this route, do you mind putting a TODO comment just so we don't forget about this?

PR #47, opened 2020-05-11
Frontend fixes

ewanmay:


src/components/Component.css
```diff
@@ -1,6 +1,6 @@
 .center-button {
   position: absolute;
-  left: 48%;
+  left: calc(50% - 90px);
```
atatmaja, line 3: why this change? did you test with different window sizes?

src/components/charts/Charts.tsx
```diff
@@ -60,35 +74,34 @@ const renderCustomizedLabel = (props: any) => {
             </div>
             <Dropdown
               placeholder='Select Independent Variable'
-              defaultValue={yAxisOptions[0].text}
-              fluid
-              selection
-              item={true}
+              defaultValue={'1'}
+              fluid selection
+              // item={true}
```
atatmaja, line 79: let's remove this comment if we're sure it won't be used

src/metaAnalysis.ts
```diff
@@ -2,6 +2,11 @@ import { AirtableRecord } from "./types"
 
 const Z_SCORE: number = 1.96;
 
+export enum aggregationFactor {
```
atatmaja, line 5: might be more consistent to put this in types.ts. not sure that's necessary tho, up to you

src/components/Component.css
ewanmay: essentially the button width is set to 180px; so it will always be centered with this calc

PR #48, opened 2020-05-13
Dashboard features

ewanmay:


src/components/map/Legend.tsx
```diff
@@ -1,47 +1,37 @@
...
+          return `<div className="col-12 p-0 flex">
+          <div className="col-12 p-0"> ${from}%${to ? "&ndash;" + to + "%" : "+"}</div>
+          <i className="legend col-12 p-0"style="background:${getColor(from)}"></i>
+          </div>`
         }
       })
-
-      div.innerHTML = labels.join("<br>");
+      labels.push(title);
+      console.log(labels);
```
atatmaja, line 33: remove console log

src/components/map/Map.tsx
```diff
@@ -45,13 +48,27 @@ export default function Map() {
     }
   }
 
+  const getLogit = (percentage: number) => {
```
atatmaja, line 51: I think it would be a good idea to abstract all this bucketing logic into its own set of functions, doesn't really have much to do with the view layer itself

atatmaja: Including this getColors function

src/components/map/Legend.tsx
```diff
@@ -1,47 +1,37 @@
...
-      const div = L.DomUtil.create("div", "info flex legend");
-
+      const div = L.DomUtil.create("div", "info flex legend center-item");
+      const title = '<h4 className="legend-title col-12 p-0 flex middle">Seroprevalence (logit scale) </h4>'
       const labels = buckets.map((value, index) => {
         if (index !== buckets.length) {
           const from = value;
           const to = buckets[index + 1]
-          return `<div className="col-12 p-0 flex"><i style="background:${getColor(from + 1)}"></i>${from}${to ? "&ndash;" + to : "+"}<div>`
+          return `<div className="col-12 p-0 flex">
```
atatmaja, line 26: should make this JSX look nicer

src/components/map/Legend.tsx
```diff
@@ -1,47 +1,37 @@
...
+          <div className="col-12 p-0"> ${from}%${to ? "&ndash;" + to + "%" : "+"}</div>
+          <i className="legend col-12 p-0"style="background:${getColor(from)}"></i>
```
atatmaja, line 28: would it be easier to just send an array of colours here instead of the getColor function? Or maybe a data structure that consolidates both buckets values and their respective colours?

src/components/map/Legend.tsx
ewanmay: this is a string tho

atatmaja: I guess semantically it's a string but functionally it's JSX. Just feel like it'll be easier to edit and read if we format it as such.

src/components/map/Legend.tsx
ewanmay: I am going to leave this for now but I'll make it a todo just because it's not super pressing

src/components/map/Map.tsx
ewanmay: Making it a todo

PR #61, opened 2020-05-17
Fixes based on dashboard feedback

ewanmay:


src/App.tsx
```diff
@@ -2,18 +2,16 @@ import React, { useContext, useEffect, useState } from "react";
 import { Link, Route, Switch, useLocation, Redirect } from "react-router-dom";
 import './App.css';
 import Icon from './assets/images/two-tone-light.svg';
-import Data from "./components/pages/Data";
+import Dashboard from "./components/pages/Data";
 import About from './components/static/About';
 import Auth from './components/static/Auth/Auth';
-import Home from './components/static/Home';
+import Data from './components/static/Home';
```
atatmaja, line 8: So as I understand, what was formerly called the Home tab will now be called the Data tab right? Is this something Rahul is expecting or nah?

src/components/charts/Charts.tsx
```diff
@@ -43,17 +52,21 @@ export default function Charts() {
...
-    const { x, y, width, height, value } = props;
+    const { x, y, width, height, value, index } = props;
+    let recordError = records[index].error;
+    const ratio = width / value;
+    const error = Array.isArray(recordError) ? recordError[1] : recordError;
+    const errorBarWidth = ratio * error;
     return (
       <g>
-        <text x={x + width + 10} y={y + height / 2} fill="#000" textAnchor="right" dominantBaseline="right">
+        <text x={x + width + 10 + (errorBarWidth as number)} y={y + height / 2 + 5} fill="#000" textAnchor="right" dominantBaseline="right">
```
atatmaja, line 62: better to type errorBarWidth in its definition or nah?

src/App.tsx
rahularoradfs: yep! discussed

src/components/map/Legend.tsx
```diff
@@ -15,20 +16,32 @@ const Legend = (props: legendProps) => {
     const control = L.control as any
     const legend = control({ position: "bottomright" });
     // apparently this is not caught in the typings so I have to any it
-    L.DomUtil.remove(legend)
     legend.onAdd = () => {
       const div = L.DomUtil.create("div", "info flex legend center-item");
-      const title = '<h4 className="legend-title col-12 p-0 flex middle">Seroprevalence (logit scale) </h4>'
+      if (isNaN(buckets[1])) {
+        return div
```
atatmaja, line 22: should we return some message instead of nothing here?

PR #67, opened 2020-05-18
Static content updates - May 17th

atatmaja:


src/components/sidebar/right-sidebar/RightSidebar.css
```diff
@@ -3,7 +3,8 @@
 }
 
 .filters-container {
-    overflow: auto;
+    overflow-y: auto;
```
rahularoradfs, line 6: WRT overflow-y: do we want to expand the size of this box? When you click on a dropdown filter (e.g., population group), a vertical scrollbar appears, presumably because the height of the filters container is limited

src/components/static/About.tsx
```diff
@@ -23,41 +23,80 @@ export default function About() {
...
     return (
         <div className="page col-12">
-            <div className="static-content">
+            <div className="static-content pb-2">
                 <h1>
-                    Scientific Lead:
+                    About SeroTracker
                 </h1>
+                <p>
+                    As the acute phase of the COVID-19 pandemic passes its peak, many countries are accelerating their investments in serological (antibody) testing. Understanding population antibody patterns and potential immunity is of great interest to clinicians, public health officials, and policymakers alike. <b>SeroTracker is a knowledge hub that tracks and synthesizes findings from SARS-CoV-2 serosurveillance efforts worldwide.</b>
```
rahularoradfs, line 31: modified to add just a littttle more selective bolding

src/components/sidebar/right-sidebar/RightSidebar.css
atatmaja: Do you mean this 
image

I think that's just default behaviour for the dropdown ui component we are using. The overflow-y just allows the whole filters section to be scrollable if the browser isn't tall enough (and I removed the scrollbar for aesthetic reasons there but we can readd it).

As for changing the dropdown component's behaviour, we can make it expand it fit, however I'm not sure that's the best user experience. Let's say the countries dropdown has like 50 options sometime in the future. If we were to expand the size of the dropdown box, it would take > the browser height to show the expanded dropdown, which seems a little weird.

PR #68, opened 2020-05-18
Study Details modal

ewanmay:


src/components/sidebar/right-sidebar/LastUpdated.tsx
```diff
@@ -17,9 +17,10 @@ export default function LastUpdated() {
               offset={10}
               position="top right"
               color="#455a64"
-              popupSize="mini"
+              tooltipHeader={"Last Updated"}
+              popupSize="small"
               size="sm"
-              tooltip={`Updated every 24 hours.`}/>
+              tooltip={'Last time the data was refreshed from the database, which gets updated every 24 hours.'}/>
```
atatmaja, line 23: Did Rahul or Jordan tell us to replace this string? Personally I think it's a bit wordy, not sure people really need/care about any info beyond what the update rate is.

src/context.tsx
```diff
@@ -28,7 +28,7 @@ const initialState: State = {
     country: new Set()
   },
   data_page_state: {
-    mapOpen: true
+    mapOpen: false
```
atatmaja, line 31: are we sure that we want the analyze tab to be open first? doesn't make that much sense to me

PR #71, opened 2020-05-20
Implement different meta-analysis modalities

rahularoradfs:
- Implemented three additional meta-analysis modalities, and re-added the fourth that Austin had previously implemented 
- Refactored code to enforce the type of `error` as `number []`, and coded all errors as asymmetric errors for consistency
- Fixed tooltips on bar charts to properly deal with asymmetric errors


src/metaAnalysis.ts
```diff
@@ -1,30 +1,115 @@
 import { AirtableRecord, AggregationFactor, AggregatedRecord } from "./types"
+import { faPassport } from "@fortawesome/free-solid-svg-icons";
```
atatmaja, line 2: unused import

src/metaAnalysis.ts
```diff
@@ -1,30 +1,115 @@
...
+
+// BROKEN NOW: CONFIDENCE INTERVAL TOOLTIPS
```
atatmaja, line 4: is this comment still accurate?

src/metaAnalysis.ts
```diff
@@ -1,30 +1,115 @@
...
+            return p; 
+        }
+        case "logit": {
+            return Math.log(p / (1 - p))
+        }
+        case "arcsin": {
+            return Math.asin(Math.sqrt(p))
+        }
+        case "double_arcsin": {
+            let n = N * p
```
atatmaja, line 20: not changing the definition of n after its declaration, so we can just make it a const

src/metaAnalysis.ts
```diff
@@ -1,30 +1,115 @@
...
+                variance = transformVariance(record.seroprevalence, record.denominator, method)!;
+                var_sum = var_sum + variance;
+                inv_var_sum = inv_var_sum + (1 / variance);
+                p_over_var_sum = p_over_var_sum + (prevalence / variance);
+                n = n + record.denominator;
+            }
+        });
+
+        pooled_p = p_over_var_sum / inv_var_sum;
+        let conf_inter = [pooled_p - Z_SCORE * Math.sqrt(var_sum), pooled_p + Z_SCORE * Math.sqrt(var_sum)]
```
atatmaja, line 101: again we can make this a const

src/metaAnalysis.ts
```diff
@@ -1,30 +1,115 @@
...
+
+        pooled_p = backTransform(pooled_p, method)!; 
+        error = [pooled_p - backTransform(conf_inter[0], method)!, backTransform(conf_inter[1], method)! - pooled_p];
     }
+
+    if (pooled_p - error[0] < 0) {error[0] = pooled_p}
```
atatmaja, line 107: nice typescript 1 line solution lol

atatmaja:
Maybe as a future step we can implement meta analysis as a class. That way we won't have to pass around state variables like the method type around a bunch of times. Also we could instantiate metaAnalysis objects with different method types in different parts of the app if we ever want to do something like that.

src/metaAnalysis.ts
rahularoradfs: no - removed

src/metaAnalysis.ts
rahularoradfs: removed

src/metaAnalysis.ts
rahularoradfs: done

src/metaAnalysis.ts
rahularoradfs: done

rahularoradfs:
refactored meta-analysis testing; fixed map tooltip for asymmetric error; added a different implementation of the double-arcsin transform and made that default

PR #73, opened 2020-05-22
Mobile Implementation

ewanmay:


src/components/charts/Charts.tsx
```diff
@@ -75,13 +76,15 @@ export default function Charts() {
     return longestWord;
   }
 
+
+  const isMobileDevice = useMediaQuery({ maxDeviceWidth: 1200 })
```
atatmaja, line 80: Would it be better to assess what type of device we're working with once on page load rather than on each component render like this?

src/components/mobile/MobileComponents.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
```
atatmaja, line 1: Doesn't look like we really need this anymore, let's get rid of it

src/components/pages/Data.tsx
```diff
@@ -1,20 +1,111 @@
...
+import React, { useState } from "react";
+import MediaQuery from 'react-responsive';
+import { Menu, Segment, Sidebar } from "semantic-ui-react";
 import CentralPiece from "../CenterComponent";
+import CountryList from "../sidebar/left-sidebar/CountryList";
 import LeftSidebar from "../sidebar/left-sidebar/LeftSidebar";
+import TotalStats from "../sidebar/left-sidebar/TotalStats";
+import Filters from "../sidebar/right-sidebar/Filters";
 import RightSidebar from "../sidebar/right-sidebar/RightSidebar";
+import MobileModules from '../mobile/MobileComponents';
```
atatmaja, line 12: remove this import, not using

PR #74, opened 2020-05-22
Fixes may 19

atatmaja:


src/components/sidebar/left-sidebar/TotalStats.tsx
```diff
@@ -34,7 +35,7 @@ export default function TotalStats() {
           <InformationIcon
             color="#468ac1"
             offset={-12}
-            tooltip="Total number of positive individuals who test positive for antibodies against SARS-CoV-2 (the virus that causes COVID-19) divided by total tests administered, using filtered data"
+            tooltip="Prevalence of antibodies against SARS-CoV-2, aggregated within the filters you have selected.\n\nNB: This should not be interpreted as a representative estimate of seroprevalence in the region of interest, especially if the aggregated studies have biased samples.\n\nPooling technique: fixed effects inverse-variance weighted aggregation of records with sample size over {MIN_DENOMINATOR}."
```
rahularoradfs, line 38: @atatmaja - this is one of the tooltips that will need a bit of work

src/components/sidebar/right-sidebar/Filters.tsx
```diff
@@ -50,31 +68,69 @@ export default function Filters() {
...
+            <div>
+              {buildSectionHeader('Geography', 'Select countries of interest.')}
+            </div>
+            <div>
+              {buildFilterDropdown('country', 'Country')}
+            </div> 
+          </div>
+          <div className="pb-1">
+            <div>
+              {buildSectionHeader('Study Information', 'Filter on study details, including source type (publication, preprint, news, or report) and study status (ongoing, completed).\n\nRisk of Bias: Reflects the extent to which the true prevalence may be different from the estimated prevalence. Estimated by SeroTracker reviewers based on the Joanna Briggs Institute critical appraisal tool for prevalence estimates.')}
```
rahularoradfs, line 98: @atatmaja the other header w newlines

src/components/sidebar/left-sidebar/TotalStats.tsx
ewanmay: did y'all check what the tooltip looks like? if you want to inject the variable you need to use backticks and dollar sign \`string like normal text because we want to showcase: ${variable}\`

PR #90, opened 2020-06-10
Meta analysis for records with seroprevalence = 0

atatmaja:
consider records reporting seroprevalence = 0 when performing meta analysis


atatmaja:
taking this PR down to include more comprehensive solutions in a separate PR as discussed this afternoon @rahularoradfs

PR #95, opened 2020-06-12
Staging

ewanmay:


src/App.tsx
```diff
@@ -22,74 +21,36 @@ function App() {
...
-  }
 
-  // AUTHENTICATION
-  // let authStatus = false;
-
-  // if (localStorage.hasOwnProperty('authenticated')) {
-  //   authStatus = localStorage.getItem('authenticated') !== null;
-  // }
-  // const [auth, setAuth] = useState(authStatus);
+    const handleResize = () => {
```
atatmaja, line 25: just curious what's the purpose of this specifically

src/components/map/Map.tsx
```diff
@@ -133,23 +84,27 @@ export default function Map() {
       fillOpacity: 0.7
     });
   };
-
+  const seroPrevalence = Translate("Seroprevalence")
```
atatmaja, line 87: why assign variables here instead of just supplying Translate("Seroprevalence") right into the JSX?

src/components/sidebar/left-sidebar/CountryList.tsx
```diff
@@ -4,30 +4,33 @@ import { AppContext } from "../../../context";
...
+        {Translate('SeroPrevalenceByCountry').toUpperCase()}
       </div>
-      <div className="item-list-container">
-        {aggregatedData.map(o => {
-          return (
-            <div className="col-12 p-0 flex item-container" key={Math.random()}>
-              <div className="p-2 country-name">
-                {o.name}
+      <div className="list">
+        {/* <div className="list-container"> */}
```
atatmaja, line 20: should we remove this commented div?

src/components/sidebar/left-sidebar/TotalStats.tsx
```diff
@@ -24,36 +25,37 @@ export default function TotalStats() {
   const countries = countryDict.length;
 
   const countriesOrCountry = () => {
```
atatmaja, line 27: I've seen a few of these general string manipulation functions pop up in our code. Would it be best to move them to some generic utils file?

Pretty minor and not at all necessary for this PR btw

src/components/sidebar/right-sidebar/Filters.tsx
```diff
@@ -1,21 +1,41 @@
...
+          formatted_options.push({
+            key: o,
+            text: getCountryName(o, state.language, jsonObjectString),
+            value: o
+          })
+        });
+        break;
+      default:
+        options.forEach((o: string) => {
+          const translatedString = Translate(jsonObjectString, [toPascalCase(o)]);
```
atatmaja, line 29: why pascal case?

ewanmay: because you need to turn them into that to access the json objects

src/components/sidebar/left-sidebar/TotalStats.tsx
ewanmay: gonna leave this for now

PR #106, opened 2020-06-21
Date slider

ewanmay:


atatmaja:
@ewanmay mind looking into this failing check before I start reviewing?

src/types.ts
```diff
@@ -20,7 +20,7 @@ export type AirtableRecord = {
     age: string[] | null,
     denominator: number | null,
     seroprevalence: number | null,
-    publish_date?: string | null,
+    publish_date?: string[] | string | null,
```
atatmaja, line 23: Are we sure it can either be a string array or a string? Double check the Airtable SOP: https://docs.google.com/spreadsheets/d/1UT0tZSj7k4OksihoIk7F3zkkoiScHzj2shkPMBF5uCQ/edit#gid=0.

Note: sometimes Airtable does this weird thing where string fields that are linked to another table show up on that other table as a single element array. If this is the case we consistently see, I would probably just map it to a string here always.

src/components/sidebar/right-sidebar/date-slider/Slider.tsx
```diff
@@ -0,0 +1,171 @@
...
+      count: totalCount,
+      label: Translate("Today")
+    }
+    const monthTicks: TickItem[] = monthRange.map(o => ({
+      tick: {
+        id: o.date.toString(),
+        value: o.date.getTime(),
+        percent: o.percentage
+      },
+      label: o.date.toLocaleDateString("en-US", formatOptions),
```
atatmaja, line 78: is this locale something we wanna switch in fr version?

src/types.ts
ewanmay: Okay, I just had some issues

src/components/sidebar/right-sidebar/date-slider/Slider.tsx
ewanmay: yep, agreed

src/types.ts
ewanmay: this one I left

PR #114, opened 2020-07-01
Consume meta analysis endpoint

atatmaja:


src/components/map/Map.tsx
```diff
@@ -20,29 +19,41 @@ export default function Map() {
...
+        setForceUpdate(Math.random())
       }
-      return { ...feature, properties: { ...feature.properties, seroprevalence: null, error: null, n: null, num_studies: null } }
-    })
-    setMapRecords(importGeo)
 
-    // we need to update the key on the GEOJSON to let react know it's time to rerender
-    setForceUpdate(Math.random())
-  }, [state.filtered_records, state.language, state.filters])
+      updateMap();
```
ewanmay, line 45: wouldn't this need to be like updateMap().Then() if it's an async method?

src/components/map/Map.tsx
```diff
@@ -20,29 +19,41 @@ export default function Map() {
...
-
-    // We will iterate through all the features in the geoJson
-    // if they are in the country dict we will attach their aggregated data to the feature for displaying
-    importGeo.features = features.map(feature => {
-      const country = prevalenceCountryDict[feature?.properties?.name];
-      if (country && country.seroprevalence) {
-        const { seroprevalence, error, n, num_studies } = country;
-        return { ...feature, properties: { ...feature.properties, seroprevalence, error, n, num_studies } }
+    if(state.country_prevalences.length > 0){
+      const updateMap = async () => {
```
ewanmay, line 23: Why is this async I don't see any awaits?

src/components/pages/Dashboard/Dashboard.tsx
```diff
@@ -1,13 +1,32 @@
...
+    if(state.filtered_records.length > 0){
+      const updateCountryPrevalence = async () => {
+        const api = new httpClient();
+        const countryPrevalences = await api.postMetaAnalysis(state.filtered_records, AggregationFactor.country);
+        dispatch({
+          type: 'UPDATE_COUNTRY_PREVALENCES',
+          payload: countryPrevalences
+        });
+      } 
+      updateCountryPrevalence();
```
ewanmay, line 26: isn't there a chance that this doesn't exist?

src/components/sidebar/left-sidebar/TotalStats.tsx
```diff
@@ -1,31 +1,30 @@
...
-    .map(o => {
-      if (o.seroprevalence !== null && o.denominator !== null) {
-        return o.country
-      }
-      return null;
-    })
-    .filter(onlyUnique);
-  const countries = countryDict.length;
+  useEffect(() => {
+    if(state.filtered_records.length > 0){
```
ewanmay, line 17: Is it smarter to constantly check inputs that we pass or the inputs that you guys receive on the backend? As a service I feel like it's smarter to sanitize your inputs rather than expect perfection

src/httpClient.ts
```diff
@@ -74,4 +99,74 @@ export default class httpClient {
             updated_at: updated_at_string
         };
     }
+
+    async postMetaAnalysis(records: AirtableRecord[], aggregation_variable: AggregationFactor, meta_analysis_technique: string = 'fixed', meta_analysis_transformation: string = 'double_arcsin_precise'){
```
ewanmay, line 103: I might want to walk through this with you, it's a little hard to read

src/components/map/Map.tsx
atatmaja: yeah will change this, sorry this was leftover from a previous implementation

src/components/map/Map.tsx
atatmaja: updateMap shouldn't be async

src/components/pages/Dashboard/Dashboard.tsx
atatmaja: chance that what doesn't exist

src/components/sidebar/left-sidebar/TotalStats.tsx
atatmaja: We actually probably don't need to do this check for state.filtered_records.length > 0

src/components/pages/Dashboard/Dashboard.tsx
ewanmay: nvm

src/components/sidebar/left-sidebar/TotalStats.tsx
ewanmay: Gotcha, okay

PR #116, opened 2020-07-02
Imperfect prelim privacy policy

ewanmay:


src/utils/translate/en.json
```diff
@@ -180,6 +181,8 @@
     "Male": "Male",
     "Unspecified": "Unspecified"
   },
+  "TermsOfUse": "Terms of Use",
+  "CookiePolicy": "CookiePolicy",
```
atatmaja, line 185: Is this supposed to be one word

PR #146, opened 2020-08-07
New popup

ewanmay:


src/utils/mapUtils.tsx
```diff
@@ -139,25 +139,38 @@ export const createPopup = (properties: any, language: LanguageType) => {
     </div>)
 }
 
+const createPopupRegionSection = (regionalEstimate: RegionalPrevalenceEstimate, title: string) => {
```
atatmaja, line 142: Minor but can we call this something else, maybe createPopupGeographySection. Just because we use this function to display the national/local/sublocal estimate info in the pop-up in addition to regional info

PR #152, opened 2020-08-27
explore & analyze seperated out into individual tabs. Data air table …

brett-stephen:
…replaced with ref table


ewanmay (approved):
If it all looks good, and you double checked:

1. Filtration when moving pages
2. Mobile view
3. Center button is gone
4. Loading times are good

Then I'm cool with merging

atatmaja (changes requested):
change base to staging plz. once that's done I'm good with it :)

brett-stephen:
changed to staging

PR #158, opened 2020-10-09
redirect dashboard to explore for bookmarked users

brett-stephen:
updated references tab:
- Make sure the analyze tab popup only appears once per "session"
- /Dashboard should redirect to /Explore


atatmaja (approved):
LGTM

PR #160, opened 2020-10-19
geojson updates to match new country mapping

rahularoradfs:


atatmaja (commented):
Looks good to me. But please ensure that applying country filter combinations that incorporate the names that you've updated works as expected. Could be a point of failure.

PR #167, opened 2020-11-07
Use calendar range picker to filter instead of slider

joanna-chen:
Using Airbnb's open-source react-dates calendar picker.

Input view:
image

Calendar view: 
image

### DIFF
--- a/src/components/sidebar/right-sidebar/CalendarPicker.tsx
+++ b/src/components/sidebar/right-sidebar/CalendarPicker.tsx
@@ -1,0 +1,110 @@
+import React, { useContext, useEffect, useState } from "react";
+import { AppContext } from '../../../context'
+import "./CalendarPicker.css"
+import Translate from "../../../utils/translate/translateService";
+
+import * as moment from 'moment';
+import 'react-dates/initialize';
+import { DateRangePicker, SingleDatePicker, DayPickerRangeController } from 'react-dates';
+import 'react-dates/lib/css/_datepicker.css';
+
+export default function CalendarPicker() {
+  const [state, dispatch] = useContext(AppContext);
+
+  const [startDate, setStartDate] = useState(null);
+  const [endDate, setEndDate] = useState(null);
+  const [focusedInput, setFocusedInput] = useState("startDate");
+
+  useEffect(() => {
+    if(state.airtable_records) {
+      const orderedDates = state.airtable_records
+      .filter(o => o.publish_date !== null && o.publish_date !== undefined)
+      .map(o => o.publish_date instanceof Array ? Date.parse(o.publish_date[0] as string) : Date.parse(o.publish_date as string))
+      .sort((a, b) => a - b);
+
+      const startDate = orderedDates[0];
+      setStartDate(moment(startDate));
+      setEndDate(moment());
+    }
+  }, [state.airtable_records])
+
+  const onDatesChange = (range: object) => {
+    const start = range.startDate;
+    const end = range.endDate;
+    if (moment.isMoment(start)) {
+      setStartDate(start);
+    }
+
+    if (moment.isMoment(end)) {
+      setEndDate(end);
+    }
+
+    if (start != null && end != null) {
+      const values = [start.valueOf(), end.valueOf()]
+      dispatch({
+        type: 'UPDATE_FILTER',
+        payload: {
+          filter_type: 'publish_date',
+          filter_value: values
+        }
+      })
+    }
+  }
+
+  const onFocusChange = (focus: object) => {
+    setFocusedInput(focus);
+  }
+
+  function isAfterDay(a, b) {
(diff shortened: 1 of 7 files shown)


src/components/sidebar/right-sidebar/RightSidebar.tsx
```diff
@@ -7,12 +8,113 @@ import "./RightSidebar.css"
...
+      .sort((a, b) => a - b);
+
+      const startDate = orderedDates[0];
+      setStartDate(moment(startDate));
+      setEndDate(moment());
+    }
+  }, [state.airtable_records])
+
+  const onDatesChange = (range: object) => {
+    console.log(range)
```
ewanmay, line 37: Probs wanna drop this

src/components/sidebar/right-sidebar/RightSidebar.tsx
```diff
@@ -7,12 +8,113 @@ import "./RightSidebar.css"
 import Translate from "../../../utils/translate/translateService";
 import { Link } from "react-router-dom";
 
+import * as moment from 'moment';
```
ewanmay, line 11: I think there is a build error because this isn't in the package.json

src/components/sidebar/right-sidebar/RightSidebar.tsx
```diff
@@ -7,12 +8,113 @@ import "./RightSidebar.css"
...
+
+  const [startDate, setStartDate] = useState(null);
+  const [endDate, setEndDate] = useState(null);
+  const [focusedInput, setFocusedInput] = useState("startDate");
+
+  useEffect(() => {
+    if(state.airtable_records) {
+      const orderedDates = state.airtable_records
+      .filter(o => o.publish_date !== null && o.publish_date !== undefined)
+      .map(o => o.publish_date instanceof Array ? Date.parse(o.publish_date[0] as string) : Date.parse(o.publish_date as string))
```
ewanmay, line 27: There might be some merge conflicts based on my rearchitecture changes, we are going to be using state.filteredRecords for now.

src/components/sidebar/right-sidebar/RightSidebar.tsx
```diff
@@ -7,12 +8,113 @@ import "./RightSidebar.css"
...
+  const onFocusChange = (focus: object) => {
+    setFocusedInput(focus);
+  }
+
+  function isAfterDay(a, b) {
+    if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
+    return !isBeforeDay(a, b) && !isSameDay(a, b);
+  }
+
+  function isBeforeDay(a, b) {
```
ewanmay, line 69: nitpick: Could be nice to make variable names a little more descriptive. 

We could also move this functionality to a separate file specifically for the datepicker just to keep things modular.

src/components/sidebar/right-sidebar/RightSidebar.tsx
```diff
@@ -7,12 +8,113 @@ import "./RightSidebar.css"
...
   return (
     <div className="justify-content-between sidebar-container">
       <div className="filters-container mb-3">
         <Filters />
-        <Slider />
+        <div className="date-range-container">
+          <div className="section-title space-between flex">
+            {Translate('DateRange')}
+          </div>
+          <DateRangePicker
```
ewanmay, line 103: We also need to implement this on our mobile views, analyzeMobile.tsx and exploreMobile.tsx

ewanmay: Because of that, I think it's a good idea to extract this functionality to it's own component for reusability

ewanmay (changes requested):

src/components/sidebar/right-sidebar/RightSidebar.tsx
joanna-chen: I think for these pure helper functions that are just comparing two moments, a and b seem reasonable but I totally get the instinctual reluctance of seeing single letter vars haha

src/components/sidebar/right-sidebar/CalendarPicker.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { useContext, useEffect, useState } from "react";
+import { AppContext } from '../../../context'
+import "./CalendarPicker.css"
+import Translate from "../../../utils/translate/translateService";
+
+import * as moment from 'moment';
```
ewanmay, line 6: We are currently using another library called date-fns which is more smaller and more frequently updated than moment. moment at this point is considered a deprecated library. 
Check this out: https://momentjs.com/docs/#/-project-status/ for more information on why not to use moment.

The library date-fns is what we are using and you can find the docs here:
https://date-fns.org/v2.16.1/docs/

It's being used in src/utils/utils.tsx. Check it out there for more info!

ewanmay (changes requested):

PR #171, opened 2020-11-13
Country name fixes

brett-stephen:
- added last ditch case to use unlocalized name if i18n module fails.
- removed unused geojson.


atatmaja (approved):
LGTM as long as you've tested to ensure no breaking changes.
Also once you merge this, can you updates staging pls.

PR #172, opened 2020-11-13
Data tab rework

brett-stephen:
Ticket: https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recEA7vJVdo0YIKWH?blocks=hide

What needs to be addressed (pulled from ticket): 
- LH sidebar featuring an updated description (TBD by Jordan/Rahul)
- Where to move "Manuscript" and "Methods" sections?


brett-stephen:
Functionality complete, ready to be reviewed/tested

brett-stephen:
> 
> 
> Ticket: https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recEA7vJVdo0YIKWH?blocks=hide
> 
> What needs to be addressed (pulled from ticket):
> 
>     * LH sidebar featuring an updated description (TBD by Jordan/Rahul)
> 
>     * Where to move "Manuscript" and "Methods" sections?

Still needs to be addressed, but may not be important at this point

src/components/charts/ReferencesSearch.tsx
```diff
@@ -0,0 +1,86 @@
...
+                                                        
+const initialState = {
+  loading: false,
+  results: [],
+  value: '',
+}
+
+const MAX_SUGGESTIONS = 5
+
+function searchReducer(state : any, action : any) {
```
atatmaja, line 15: good idea! maybe we want to abstract the reducer logic into its own file

src/components/charts/ReferencesTable.tsx
```diff
@@ -18,6 +22,7 @@ export default function ReferencesTable() {
   const [pageLength, setPageLength] = useState(5);
   const [column, setColumn] = useState('denominator');
   const [direction, setDirection] = useState('descending');
+  const [searchResults, setSearchResults] = useState<AirtableRecord[] | undefined>(undefined);
```
atatmaja, line 25: Couldn't the initial value for searchResults just be an empty array instead of undefined? That would simplify typing

atatmaja (commented):
Generally looking good! Few minor suggestions.

Also:
- Would be awesome if you could add screenshots of the new views in action to this PR
- Not sure what the best path forward with merging this is (whether we get this out to master first or ship it with the re-architecture and merge to staging for now), @ewanmay thoughts?

src/components/charts/ReferencesSearch.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [state, dispatch] = React.useReducer(searchReducer, initialState)
+  const { loading, results, value } = state
+
+  const timeoutRef : any = React.useRef()
+
+  const handleSearchChange = React.useCallback((e, data) => {
+    clearTimeout(timeoutRef.current)
+    dispatch({ type: 'START_SEARCH', query: data.value })
+
+    timeoutRef.current = setTimeout(() => {
```
atatmaja, line 49: why do we need a timeout here? can searches really take that long?

src/components/charts/ReferencesTable.tsx
brett-stephen: There is a slight difference between an empty array of records vs undefined. Undefined means that no search has been made, the search query is empty. An empty array of records means that a search was made but no results were found. We show the full set of airtable records by default when the searchResult is undefined, and nothing when the records array is empty

src/components/charts/ReferencesSearch.tsx
brett-stephen: This timeout is used for executing the search, so after 800 milliseconds since the users last keystroke, the search will be executed. I'm just following semantic-ui's design for this one

src/components/charts/ReferencesSearch.tsx
brett-stephen: I can do this 👍

brett-stephen:
closed as rework was merged with staging

PR #184, opened 2020-12-05
Restrict fields retrieved on explore tab

atatmaja:


src/httpClient.ts
```diff
@@ -65,6 +65,7 @@ export default class httpClient {
     }
     
     async getAirtableRecords(filters: Filters,
+        only_explore_columns=false,
```
ewanmay, line 68: underscore king over here, I do want to start enforcing camel casing on the frontend

ewanmay (approved):
looks fine to me

PR #185, opened 2020-12-08
The Biggest Datepicker Pull in the history of ever

ewanmay:
This is actually a rearchitecture of our internal state. 

1) We now pass a prop around to determine which part of the state we look into to retrieve cached data
2) Created utils to manage the updating of state in a central place
3) Now caching estimateGradePrevalences
4) Now caching meta analyzed records
5) References table now gets all data and displays it based off it's own call as opposed to filtered records from other parts of the state.
6) Explore is open is now set based on navigation to and from explore tab
7) Created a call to populate the state at app start.
8) Implemented methods in our reducer to accommodate horizontal state scaling in the future
9) The only places that control updates now are the filters, and the 1 call made by App.tsx to populate. 
10) Changed the time name of estimate grades from AlternateAggregateRecord to EstimateGradePrevalence.
11) Implement a loader for both analyze and explore


src/components/charts/Charts.tsx
```diff
@@ -17,10 +17,9 @@ export default function Charts() {
   const [yAxisSelection, setYAxis] = useState(AggregationFactor.country);
   const isMobileDeviceOrTablet = useMediaQuery({ maxWidth: mobileDeviceOrTabletWidth })
   const [state, dispatch] = useContext(AppContext);
-  const { filters, showAnalyzePopup } = state;
-  const initState = [] as AggregatedRecord[];
+  const { showAnalyzePopup } = state;
   const [showModal, toggleModal] = useState(true);
-  const [records, setRecords] = useState(initState);
+  const records = state.analyze.metaAnalyzedRecords;
```
atatmaja, line 22: do we even need records here? if it's getting pulled in the references tab anyways

src/components/mobile/AnalyzeMobile.tsx
```diff
@@ -66,7 +68,7 @@ export default function AnalyzeMobile() {
             size={"lg"} />
         </Sidebar>
         <Sidebar.Pusher className="fill flex">
-        <Charts />
+          {/* <Charts /> */}
```
atatmaja, line 71: let's leave this in for now. will need to rework/investigate further but afaik the analyze tab on prod mobile is just as bad as staging mobile

src/components/pages/Dashboard/Explore.tsx
```diff
@@ -1,38 +1,58 @@
...
 export default function Explore() {
   const isMobileDeviceOrTablet = useMediaQuery({ maxDeviceWidth: mobileDeviceOrTabletWidth });
+  const [state, dispatch] = useContext(AppContext)
+
+  useEffect(() => {
+    dispatch({
+      type: "UPDATE_EXPLORE_IS_OPEN",
+      payload: true
+    })
+    return () => {
```
atatmaja, line 22: what's the point of this here?

src/components/pages/Data.tsx
```diff
@@ -28,6 +25,13 @@ export default function Data() {
         })
     }
 
+    const getAllRecords = async () => {
+        const api = new httpClient()
+        // TODO: Figure out a better place to put this so we don't keep updating this either
```
atatmaja, line 30: could have a "data" section in the context as well, we may eventually put filters in the data tab. that can happen after release tho

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    if (isStart) {
+      setStartDate(date);
+    }
+    else {
+      setEndDate(date);
+    }
+    await updateFilters(dispatch,
+      pageState.filters,
+      "publish_date" as FilterType,
```
atatmaja, line 35: I think we're filtering by sampling_end_date now?

atatmaja: Don't think this nomenclature affects anything (since the backend is responsible for the actual filtering), but maybe we want to rename for consistency

src/httpClient.ts
```diff
@@ -59,15 +59,15 @@ export default class httpClient {
         // We know that only these 3 isotypes will ever be reported, thus we can hardcode
         options.isotypes_reported = new Set(["IgG", "IgA", "IgM"]);
         const updatedAt = format(parseISO(response.updated_at), "dd-MM-yyyy");
-        const maxDate = parseISO(response.min_date);     
-        const minDate = parseISO(response.max_date);
+        const maxDate = parseISO(response.max_date);
```
atatmaja, line 62: lmao nice catch

src/utils/stateUpdateUtils.tsx
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  const updatedFilters: Filters = Object.assign({}, filters);
+  updatedFilters[filterType] = new Set(filterValue);
+
+  const api = new httpClient()
+  // Update current records when called
+  const [records,
+    estimateGradePrevalences, reAggregatedRecords] = await
+      Promise.all([
```
atatmaja, line 38: nice

src/utils/stateUpdateUtils.tsx
```diff
@@ -0,0 +1,120 @@
...
+        api.getAirtableRecords(filters, exploreIsOpen),
+        api.getEstimateGrades(filters),
```
atatmaja, line 40: we only need estimate grades when page === "explore" right?

src/utils/stateUpdateUtils.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const api = new httpClient()
+
+  dispatch({
+    type: "CHANGE_LOADING",
+    payload: {
+      pageStateEnum: page,
+      isLoading: true
+    }
+  })
+  // const records = await api.getAirtableRecords(filters, exploreIsOpen)
```
atatmaja, line 83: can we remove these comments?

src/components/pages/Dashboard/Analyze.tsx
```diff
@@ -1,38 +1,43 @@
...
-    </React.Fragment>
+    <>
+      <div className="fill flex dashboard">
+        {!isMobileDeviceOrTablet ?
+          (<div className="fill flex">
+            <div className="col-2 p-0 flex">
+              <LeftSidebar page={PageStateEnum.analyze} />
+            </div>
+            <div className="col-8 p-0 flex">
+              <Loader indeterminate active={state.analyze.isLoading}></Loader>
```
atatmaja, line 27: nice

src/utils/stateUpdateUtils.tsx
ewanmay: estimate grades are used for the summary stats

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
ewanmay: We can do that yeah, just didn't worry about it for now

src/components/pages/Dashboard/Explore.tsx
ewanmay: it's a clean up function that runs when the component dismounts from the DOM

src/components/pages/Data.tsx
ewanmay: I'm just gonna leave this here for now because of how many records we would have to store in context that would otherwise bog down the app

src/components/charts/Charts.tsx
ewanmay: we pass records into the bar chart

atatmaja: yeah whoops I'm dumb, I misread this as like individual records rather than meta analyzed

atatmaja (approved):

PR #187, opened 2020-12-08
Minor datepicker fixes

atatmaja:
Update datepicker UI and verbiage
image

Fix explore spinner loading


atatmaja:
Merging for now so that we can view the staging site, may revert

PR #191, opened 2020-12-15
Who map compliance

brett-stephen:

PR #192, opened 2020-12-22
Estimate pins view

joanna-chen:
Estimate pins added to the Explore tab: 
image

A popup comes up when a pin is pressed: 
image

Country hover popup is changed to keep consistency in styling: 
image

Note: this will be merged after #191 is merged first


brett-stephen (approved):
It looks good 👍

src/components/map/Legend.tsx
```diff
@@ -28,11 +34,34 @@ export default function Legend(props: LegendProps) {
   //   return null;
   // })
 
+  const clickEstimatePinsCheckbox = (e: React.MouseEvent<HTMLElement>) => {
+    dispatch({ type: 'TOGGLE_ESTIMATE_PINS' })
+    console.log(state.showEstimatePins)
```
atatmaja, line 39: remove console log

src/components/map/Map.tsx
```diff
@@ -78,6 +103,66 @@ export default function Map() {
...
+                    Study name
+                  </div>
+                  <div className="popup-text">
+                    {`${record.source_name}`}
+                  </div>
+                  <div className="popup-heading">
+                    Location
+                  </div>
+                  <div className="popup-text">
+                    {record.city && `${record.city}`}{record.city && record.city.length !== 0 && record.state && record.state.length !== 0 && ", "}{record.state && `${record.state}`}{record.state && record.state.length !== 0 && record.country && ", "}{record.city && record.city.length !== 0 && record.state && record.state.length === 0 && record.country && ", "}{record.country && `${record.country}`}
```
atatmaja, line 134: I think we might already have similar functionality here to compute "location text" (see https://github.com/serotracker/sero-can-webapp/blob/master/src/components/shared/StudiesTable.tsx#L66)

Let's consolidate both into a shared utility function for improved consistency/maintainability

src/components/map/Map.tsx
```diff
@@ -78,6 +103,66 @@ export default function Map() {
...
+                    N
+                  </div>
+                  <div className="popup-text">
+                    {`${record.denominator}`}
+                  </div>
+                  <div className="popup-heading">
+                    Population group
+                  </div>
+                  <div className="popup-text">
+                    {record.population_group ? `${record.population_group}` : "N/A"}
```
atatmaja, line 152: Default should be "Not Reported", also let's use translations here pls

src/components/map/Map.tsx
```diff
@@ -78,6 +103,66 @@ export default function Map() {
...
+                eventHandlers={{
+                  click: (e) => {
+                    (e.target as any).openPopup();
+                  },
+                }}
+                icon={record.estimate_grade === "National" ? nationalMarker : (record.estimate_grade === "Regional" ? regionalMarker : localMarker)}
+                key={`marker-${idx}`} 
+                position={[record.pin_latitude, record.pin_longitude]}
+              >
+                <Popup autoClose={false} className="pin-popup">
```
atatmaja, line 120: In general let's use our translation system here instead of hardcoded strings

src/tests/testData.ts
```diff
@@ -13,6 +13,9 @@ export const testRecords: AirtableRecord[] = [
         denominator: 2000,
```
atatmaja, line 13: @ewanmay @brett-stephen are we even using this test data anymore lol. Let's make a task to either incorporate these into proper tests or get rid of them if we don't have any immediate plans to do this. Seems like a waste of time to be maintaining this.

atatmaja (changes requested):
Good work @joanna-chen! Just a few minor comments, mostly around using translations instead of hardcoding strings.

src/components/map/Map.tsx
atatmaja: Actually, would it be easier for you to deal with incorporating translations here @ewanmay given that you're currently working on translations as a whole?

ewanmay: Yeah I can do that

atatmaja (approved):
LGTM! @ewanmay can you take a look today and we can merge if it's all good

PR #195, opened 2021-01-10
French Translations

ewanmay:
Actions:
- Made sure all strings were translated
- Added translations we were missing or the ones that were incorrect
- Added a url so you could click on the link for the study pins


src/utils/translate/en.json
```diff
@@ -2,22 +2,18 @@
   "95%ConfidenceInterval": "95% Confidence Interval",
```
atatmaja, line 2: What's the purpose of leaving this old en.json file?

atatmaja (commented):
Looks fine to me once you answer my question

atatmaja:
Oh also we should find a place to commit any of the scripts you used to update the language json files @ewanmay

src/utils/translate/en.json
ewanmay: Because you need it to hold the english strings

ewanmay:
> commit any of the scripts you used to update the language json files

Did not use any scripts aside from the one you used to get things started

atatmaja:
> > commit any of the scripts you used to update the language json files
> 
> Did not use any scripts aside from the one you used to get things started

Did you account for new strings that were added after the translation spreadsheet was made? If so, I guess you went through the spreadsheets manually?

src/utils/translate/en.json
atatmaja: Uhh but we also have another file at path `src/utils/translate/newjson/en.json`, what's the distinction here

ewanmay: that was not supposed to be there, I have deleted it

ewanmay:
> > > commit any of the scripts you used to update the language json files
> > 
> > 
> > Did not use any scripts aside from the one you used to get things started
> 
> Did you account for new strings that were added after the translation spreadsheet was made? If so, I guess you went through the spreadsheets manually?

Yes I did, I went through them manually

atatmaja (approved):

PR #197, opened 2021-01-16
Add sampling dates to pin popup

joanna-chen:
image (1)


src/components/map/Map.tsx
```diff
@@ -135,6 +135,12 @@ export default function Map() {
                   <div className="popup-text">
                     { getGeography(record.city, record.state, record.country) }
                   </div>
+                  <div className="popup-heading">
```
atatmaja, line 138: We should render this conditionally (only if record.sampling_start_date & record.sampling_end_date are not NULL)

atatmaja (approved):
LGTM

PR #203, opened 2021-01-30
Place links on insights tab into carousels

joanna-chen:
## Describe the feature or bug that this PR addresses.
- Put links from insights tab into carousels 
- Added a media section on the insights tab

## If applicable, include screenshots of the feature/bugfix introduced by this PR.
image

- Locally deployed and tested the carousel works as expected
- Crosschecked with the insights spreadsheet that all the info is accurate

## Does this PR depend on any recent backend work?
Nope

Possible TODO: we might want to also change the articles section to use the same mapping into InsightsCard flow that I introduce in this PR


atatmaja:
Nice work! I have 2 requests:

1. Can we put a bit of padding between each card in the carousel
2. Can we add an image for each article in the media carousel, just take the first image presented in the article if applicable.

rahularoradfs:
This looks great, Joanna!! Any chance we could create separate sections for research, reports, and media - that is, split research articles and reports into two different sections?

rahularoradfs:
Would also be great to get images up for media tab at some point, but definitely not essential for this ticket :)

atatmaja:
Okay what rahul suggested for next steps for this PR sounds good. Ping me when it's ready for another review @joanna-chen

src/components/pages/insights/InsightsCard.tsx
```diff
@@ -51,3 +63,277 @@ export default function InsightsCard(props: InsightsCardProps) {
...
+    title: Translate('IndustryReportTitles', ['PrePrint']),
+    url: "https://www.medrxiv.org/content/10.1101/2020.05.10.20097451v1"
+  },
+]
+
+const listOfMediaInsightProps: InsightsCardProps[] = [
+  {
+    identifier: Translate("Media"),
+    date: "May 23, 2020",
+    title: "Understanding the Virus and Its Unanswered Questions",
```
atatmaja, line 155: Let's get @simonarocco to help us translate these titles before releasing

atatmaja (approved):
Code LGTM, but we should arrange to get these strings translated before we release

src/utils/translate/fr.json
```diff
@@ -168,6 +168,26 @@
     "Geography": "Géographie",
     "GeographyTooltip": "Sélectionner les pays concernés.",
     "Here": "Ici",
+    "IndustryMediaTitles": {
+      "UnderstandingVirus": "Comprendre le virus et ses questions sans réponse",
```
simonarocco, line 172: @atatmaja @joanna-chen I feel like we should leave these titles in English. Their official title is in English and you wouldnt translate a journal article title or a book title for example into French if you were speaking about it, and you wouldn't translate the article itself if you were citing it. I think it's weird to translate because some of the translations don't literally map to french well so we'd have to basically say it in a different way in French and that would take a bit of time. Also the translation on line 178 is from Spanish so it didn't translate properly to French

simonarocco (commented):
see comment

src/utils/translate/fr.json
jordanvanwyk: can push without translations just need to flag it here: https://docs.google.com/spreadsheets/d/1lk8ps_cKIBjFy6rVMgzHkrBJVfwKamR76bJK4ZfHmH4/edit#gid=1903680511

jordanvanwyk: > Nice work! I have 2 requests:
> 
> 1. Can we put a bit of padding between each card in the carousel
> 2. Can we add an image for each article in the media carousel, just take the first image presented in the article if applicable.

Can we do articles, reports, media for the order please? 
Would be great to have fixed heights for the cards.  
Can remove the "Type" heading labeling as article for ex. since they are segregated. 
Would put all text bellow the image, Title and a smaller date bellow that.
**Title**
Also would think about doing columns instead of carosel.

atatmaja:
Tested on mobile and there still seem to be some kinks to work out. Closing for now.

PR #215, opened 2021-03-13
Cleaned up React Context

annoyingpr:

Cleaned up some of the unused variables in React Context. Data page state variables must be added when filters are added to that page. 

https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recjrEU9Ba07eYuEe?blocks=hide

No UI changes, everything should work normally. 

Just tested around the website locally to see if any functionality was broken.

Nope.

### DIFF
--- a/src/context.tsx
+++ b/src/context.tsx
@@ -7,11 +7,13 @@ import {
   PageState,
   State,
 } from "./types";
-import Translate from "./utils/translate/translateService";
 
 export const AppContext = createContext(
   {} as [State, Dispatch<Record<string, any>>]
 );
+
+const minDate = new Date(2019, 1, 1, 1);
+const maxDate = new Date();
 
 export function getEmptyFilters(): Filters {
   return {
@@ -24,35 +26,8 @@ export function getEmptyFilters(): Filters {
     overall_risk_of_bias: new Set(),
     isotypes_reported: new Set(),
     specimen_type: new Set(),
-    publish_date: new Set([new Date(2019, 1, 1, 1), new Date()]),
+    publish_date: new Set([minDate, maxDate]),
     estimate_grade: new Set(),
-  };
-}
-
-export function getDefaultFilters(): Filters {
-  return {
-    source_type: new Set([
-      Translate("SourceTypeOptions", ["Preprint"]),
-      Translate("SourceTypeOptions", ["Publication"]),
-      Translate("SourceTypeOptions", ["InstitutionalReport"]),
-    ]),
-    test_type: new Set(),
-    country: new Set(),
-    population_group: new Set([
-      Translate("PopulationGroupOptions", ["GeneralPopulation"]),
-      Translate("PopulationGroupOptions", ["BloodDonors"]),
-      Translate("PopulationGroupOptions", ["ResidualSera"]),
-    ]),
-    sex: new Set(),
-    age: new Set(),
-    overall_risk_of_bias: new Set(),
-    isotypes_reported: new Set(),
-    specimen_type: new Set(),
-    publish_date: new Set([new Date(2019, 1, 1, 1), new Date()]),
-    estimate_grade: new Set([
-      Translate("EstimateGradeOptions", ["National"]),
-      Translate("EstimateGradeOptions", ["Regional"]),
-    ]),
   };
 }
 
@@ -75,8 +50,8 @@ const initialState: State = {
   },
   calendarStartDates: {
     // Important, the fact that we use an hour here tells us that we are using a default value
-    minDate: new Date(2019, 1, 1, 1),
-    maxDate: new Date(),
+    minDate,
+    maxDate,
   },
   language: LanguageType.english,
   updatedAt: "",
--- a/src/context.tsx
+++ b/src/context.tsx
@@ -12,8 +12,8 @@ export const AppContext = createContext(
   {} as [State, Dispatch<Record<string, any>>]
 );
 
-const minDate = new Date(2019, 1, 1, 1);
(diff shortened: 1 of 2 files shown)


src/context.tsx
```diff
@@ -41,147 +50,124 @@ export function getDefaultFilters(): Filters {
...
-  },
   allFilterOptions: getEmptyFilters(),
   dataPageState: {
     exploreIsOpen: true,
-    showStudiesModal: false,
-    routingOccurred: false
   },
   calendarStartDates: {
     // Important, the fact that we use an hour here tells us that we are using a default value
     minDate: new Date(2019, 1, 1, 1),
```
annoyingpr, line 78: Is it a bad idea to use the `minDate` and `maxDate` from context for `publish_date` in these two functions:

**getEmptyFilters()**
image

**getDefaultFilters()**
image

atatmaja: I'm not opposed, makes it a bit more readable

src/context.tsx
```diff
@@ -16,23 +25,23 @@ export function getEmptyFilters(): Filters {
     isotypes_reported: new Set(),
     specimen_type: new Set(),
     publish_date: new Set([new Date(2019, 1, 1, 1), new Date()]),
-    estimate_grade: new Set()
-  }
+    estimate_grade: new Set(),
+  };
 }
 
 export function getDefaultFilters(): Filters {
```
atatmaja, line 32: might need to double check but I don't think we need getDefaultFilters anymore as this was for the analyze tab

src/context.tsx
```diff
@@ -190,22 +176,27 @@ const reducer = (state: State, action: Record<string, any>): State => {
       const pageState = newState[pageStateEnum as keyof State] as PageState;
       pageState.isLoading = isLoading;
       return newState;
-    };
+    }
     case "UPDATE_AGGREGATION_FACTOR":
       return { ...state, chartAggregationFactor: action.payload };
     case "UPDATE_EXPLORE_IS_OPEN":
```
atatmaja, line 182: Could be wise to address in another PR but this whole system around our `dataPageState` object is very outdated. We should update this logic to be simpler and tie up loose ends eventually. Maybe we have a quick sync about this sometime @brett-stephen @annoyingpr

atatmaja (commented):
A few comments, thanks for doing this clean up!

src/context.tsx
brett-stephen: Agreed. Also, a bit unrelated but it would also be nice to address the issue of the API calls saved in the app context. If you recall, we didn't have a way to distinguish between a failed API call and one that hasn't returned anything yet.

brett-stephen:
Once Austins concern has been addressed it should be good to merge. Looks good!

brett-stephen (commented):
Once Austins concern has been addressed it should be good to merge. Looks good!

src/context.tsx
annoyingpr: Sounds good, do you think it should be a part of this [ticket](https://airtable.com/tbli2lWQHAqBa6ZcI/viw8Ro4zYPYcBGXRw/recNqmd8naERYKBio?blocks=hide) or should we create a new one?

src/context.tsx
annoyingpr: Great catch!

src/context.tsx
```diff
@@ -1,8 +1,19 @@
...
+  PageState,
+  State,
+} from "./types";
 
-export const AppContext = createContext({} as [State, Dispatch<Record<string, any>>]);
+export const AppContext = createContext(
+  {} as [State, Dispatch<Record<string, any>>]
+);
+
+const minDate = new Date(2019, 1, 1, 1);
```
annoyingpr, line 15: Lemme know if you guys don't like this approach - I had to do something like this because of the circular dependencies between `initialState` and `getEmptyFilters`.

src/context.tsx
atatmaja: make new one pls

src/context.tsx
```diff
@@ -15,173 +26,123 @@ export function getEmptyFilters(): Filters {
...
       };
     case "UPDATE_FILTER": {
       const { pageStateEnum, filterType, filterValue } = action.payload;
       const newState = { ...state };
       const pageState = newState[pageStateEnum as keyof State] as PageState;
       pageState.filters[filterType as FilterType] = new Set(filterValue);
       return newState;
-    };
+    }
     // Note: deprecated but leaving here in case we ever need again
```
atatmaja, line 132: on second thought what do we think about just deleting this lol. i think I made a bad call leaving here and putting a deprecated comment

src/context.tsx
atatmaja: eh readable enough for now

atatmaja, line 15: Let's make a small ticket to make this min date reflective of our database state (earliest date in our database). Not a huge priority imo but would be nice @annoyingpr

atatmaja (approved):
Looks good as long as we've tested thoroughly.

Few action items:
- Pls make tickets that I suggested
- Delete the UPDATE_META_ANALYSIS reducer function and do a quick smoke test to make sure nothing breaks

Thanks for the work!

src/context.tsx
atatmaja: Also nit: what do we think about renaming these consts to `initialMaxDate` and `initialMinDate` to make sure these are the initial bounds on the calendar

src/context.tsx
annoyingpr: like it was never there!

PR #217, opened 2021-03-18
Fix details modal

atatmaja:

- There were several fields missing from our study details modal. This PR re-adds them.
- Removed unused test data
- Refactored http client 

image

Local smoke testing on the details modal on the data table and general regression testing

https://github.com/serotracker/iit-backend/pull/178


src/httpClient.ts
```diff
@@ -129,27 +128,24 @@ export default class httpClient {
...
             sampling_start_date: startDate,
             sampling_end_date: endDate,
             sorting_key: sorting_key,
             reverse: reverse,
             per_page: null,
             page_index: null,
-            columns: only_explore_columns ? explore_columns : all_columns,
         }
+
+        if( only_explore_columns ){
```
annoyingpr, line 145: Why this instead of `columns: only_explore_columns ? explore_columns : all_columns,`?

src/httpClient.ts
```diff
@@ -174,14 +169,18 @@ export default class httpClient {
...
                 source_type: item.source_type,
                 specimen_type: Array.isArray(item.specimen_type) ? item.specimen_type : [item.specimen_type],
                 state: item.state,
                 study_type: item.study_type,
-                test_type: item.study_type,
+                summary: item.summary,
+                test_type: item.test_type,
+                test_manufacturer: item.test_manufacturer,
                 url: item.url
             };
```
annoyingpr, line 185: Is it possible to use object destructuring for this? Check out these [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

annoyingpr (approved):
🚀  - great work, just a couple of comments!

src/httpClient.ts
atatmaja: If you don't supply the "columns" parameter the backend will supply all columns by default.

It's better if we don't have to update the supplied columns in both the front end and the backend, I think that was the reason why the study details modal got out of sync in the first place.

src/httpClient.ts
atatmaja: Typescript will make this a bit tricky I feel. I think a better long term solution is to find a proper way to map an http response to a typescript object without being so verbose. I will note this in a TODO comment.

PR #241, opened 2021-05-18
Adding vaccination policy to study modal

simonarocco:
Adding a new section to the study modal that shows the status of vaccine rollout at midpoint of study

https://airtable.com/tbli2lWQHAqBa6ZcI/viwDpHWCgFagItSc6/recKqJTkraMI2ETTT?blocks=hide

image

- Ran the frontend locally and tested that for all the types of vaccination policies the correct descriptions are shown

- No

- Yes


src/components/map/Popups/StudyPopup.tsx
```diff
@@ -1,9 +1,22 @@
 import React from "react";
+import { Button } from 'semantic-ui-react'
 import { AirtableRecord } from "types";
 import Translate from 'utils/translate/translateService';
 import { getGeography, getPossibleNullDateString } from 'utils/utils';
+import InformationIcon from "../../shared/InformationIcon";
 
 export default function StudyPopup(record: AirtableRecord) {
+    const getVaccinationIconText = (vaccinationPolicy: number) => {
+        const vaccinationPolicyMapping: Record<number, string> = {0: "No availability",
```
atatmaja, line 10: Need to incorporate this into our translation system

src/httpClient.ts
```diff
@@ -58,51 +58,12 @@ export default class httpClient {
...
-
     async getRecordDetails(source_id : string) {
         const response = await this.httpGet(`/data_provider/record_details/${source_id}`, true);
         if (!response) {
             return null;
         }
-
-        const record: AirtableRecord = this.responseToRecord(response);
-        
+        const record: AirtableRecord = response as AirtableRecord;
```
atatmaja, line 66: niceeeeee

src/components/map/Popups/StudyPopup.tsx
```diff
@@ -1,9 +1,22 @@
 import React from "react";
+import { Button } from 'semantic-ui-react'
```
atatmaja, line 2: Remove this import as we are no longer using this

atatmaja (commented):
A few comments, great first PR!

src/components/map/Popups/StudyPopup.tsx
```diff
@@ -20,6 +33,12 @@ export default function StudyPopup(record: AirtableRecord) {
         <div className="popup-text">
             {getGeography(record.city, record.state, record.country)}
         </div>
+        <div className="popup-heading">
+            {Translate("PopulationGroup")}
+        </div>
+        <div className="popup-text">
+            {record.population_group ? `${record.population_group}` : Translate("NotReported")}
```
annoyingpr, line 40: Think this should work since `record.population_group` is a string already! If it's null, it'll return the second option - check [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) out for more info
```suggestion
            {record.population_group ?? Translate("NotReported")}
```

src/components/map/Popups/StudyPopup.tsx
```diff
@@ -43,10 +62,10 @@ export default function StudyPopup(record: AirtableRecord) {
             {`${record.denominator}`}
         </div>
         <div className="popup-heading">
-            {Translate("PopulationGroup")}
+            {Translate("VaccineRolloutStatus")}
         </div>
         <div className="popup-text">
-            {record.population_group ? `${record.population_group}` : Translate("NotReported")}
+            {(record.vaccination_policy == 0) ? "Vaccines unavailable" : (record.vaccination_policy == 1 || record.vaccination_policy == 2 || record.vaccination_policy == 3) ? "Vaccines available to certain groups" : "Vaccines available to broader population"}
```
annoyingpr, line 68: Doing this out of JSX and in a variable above could make this more clear / readable

src/components/shared/InformationIcon.tsx
```diff
@@ -14,8 +14,10 @@ interface InformationIconProps {
   popupSize?: "mini" | "tiny" | "small" | "large" | "huge";
 }
 
+
```
annoyingpr, line 17: nit: Don't need these spacing changes

src/types.ts
```diff
@@ -2,35 +2,36 @@ export type AirtableRecord = {
...
-  sensitivity?: number | null;
-  specificity?: number | null;
+  test_manufacturer: string[] | null;
+  sensitivity: number | null;
+  specificity: number | null;
   country: string | null;
-  state?: string[] | null;
-  city?: string[] | null;
+  state: string[] | null;
+  city: string[] | null;
```
annoyingpr, line 15: These aren't optional anymore?

annoyingpr (approved):
Few nits, but great work!!! 🚀

src/components/map/Popups/StudyPopup.tsx
simonarocco: this makes sense to me as long as we are sure population group will be a string @atatmaja thoughts?

src/components/map/Popups/StudyPopup.tsx
simonarocco: you mean turn this into a set of if statements? what do you mean by extracting into another variable and what do you mean by magic string?

src/types.ts
simonarocco: no... @atatmaja do you remember the reason why? i forget.

src/components/map/Popups/StudyPopup.tsx
atatmaja: `population_group` is typed as a string so it will fail at the ts compiler otherwise anyways, so I am fine with this approach

src/components/map/Popups/StudyPopup.tsx
atatmaja: These strings should be translated!!

src/types.ts
atatmaja: Our typing was a bit confusing here, we made the fields optional but also typed them as `string[] | null`. I actually think the best solution is to remove the | null part and just make these optional but haven't tested enough to kniow if that will break anything.

@annoyingpr thoughts??

src/components/map/Popups/StudyPopup.tsx
simonarocco: I'm pretty sure we actually aren't using this function at all anymore so im going to remove

src/components/map/Popups/StudyPopup.tsx
simonarocco: done

src/types.ts
annoyingpr: yeah I think making them optional should provide the same function as `string[] | null`!

atatmaja (approved):
lgtm

PR #247, opened 2021-06-05
Country popup navigation

brett-stephen:
* Add partnerships drop down to nav bar
* Add mapbox map to Canada studies page
* Move canada page into /Studies/
* Reworked map to handle different configurations
* Small improvements to map usability and map popup changes
https://airtable.com/tbli2lWQHAqBa6ZcI/viwKscR1ZvgkUxVtA/recnIYHmHBaovZP9P?blocks=hide

Navbar drop down:
image

Studies Page (Canada view):
image

Improved pin popup (arrow readded):
image

It will require french translations, will get them added before we merge to main


brett-stephen:
Will probably need to go through a couple of iterations of feedback but functionality is all here.

src/App.tsx
```diff
@@ -127,19 +128,8 @@ function App() {
...
-            desktopOptions={{
-              width: "80vw",
-              height: "4100px"
-            }}
-            mobileOptions={{
-              width: "90vw",
-              height: "2800px"
-            }}
-          />
+        <Route path="/:language/Studies/:name">
```
atatmaja, line 131: I think "/Partnerships" would be more suitable for this. What do you think @jordanvanwyk @rahularoradfs

src/PartnershipsConfig.ts
```diff
@@ -0,0 +1,25 @@
+const PartnershipsConfig = {
+    Country: [
```
atatmaja, line 2: Better to make this a dictionary with key equal to the ISO3 code so we don't need to search through an array?

i.e.

```
Country: {
   "CAN": {
       "tableauKey": "CanadianTableau",
       ...
   }
}
```

atatmaja: Nit: Could create a type for this config object, would serve as good documentation for what should be supplied when creating a new config

src/components/map/Layers/Countries.tsx
```diff
@@ -1,50 +1,85 @@
...
+        source: COUNTRY_LAYER_ID,
+        sourceLayer: COUNTRY_LAYER_ID
+        });
+
     estimateGradePrevalences.forEach((country: EstimateGradePrevalence) => {
         if (country && country.testsAdministered && country.alpha3Code) {
+
+            let partnershipConfig: any = PartnershipsConfig.Country.find(x => x.iso3 === country.alpha3Code)
+
+            //PartnershipsConfig.Country.hasOwnProperty(country.alpha3Code)
```
atatmaja, line 25: remove comment

src/components/map/Layers/Countries.tsx
```diff
@@ -1,50 +1,85 @@
...
                     regionalEstimate: country.regionalEstimate,
                     sublocalEstimate: country.sublocalEstimate,
+                    partnershipConfig: partnershipConfig
                 }
             );
         }
     });
 }
 
+function SetStudiesMode(map: mapboxgl.Map, iso3Name: string) {
```
atatmaja, line 50: This function is invoked when we hover over a country right? 

I think naming is a bit misleading, let's call this `HighlightCountry` or something

src/components/map/Layers/Countries.tsx
```diff
@@ -54,80 +89,107 @@ function FilterCountryEstimates(map: mapboxgl.Map, estimateGradePrevalences: Est
...
-            map.on('sourcedata', function (e: any) {
-                if (e.sourceId === 'Countries' && map.isSourceLoaded('Countries')) {
+
+        if (map)
+        {
+            if(estimateGradePrevalences)
+            {
+                if (map?.getSource(COUNTRY_LAYER_ID) === undefined) 
+                {
+                    map.once('sourcedata', function (e: any) {
```
atatmaja, line 108: So it looks like this logic here tries to wait for 'sourcedata' to return a sourceId twice. This seems a bit brittle to me (what if we don't find e.sourceId on the third query as well?). 

Would it be advisable for us to refactor this logic to somehow keep querying for sourcedata until we get e.sourceId?

src/components/map/Layers/Countries.tsx
```diff
@@ -54,80 +89,107 @@ function FilterCountryEstimates(map: mapboxgl.Map, estimateGradePrevalences: Est
...
                     countryPop
-                    .setLngLat(e.lngLat)
-                    .setHTML(ReactDOMServer.renderToString(CountryPopup(e.features[0], state.language)))
-                    setCurrentFeatureId(f.id);
-                }
+                      .setHTML(ReactDOMServer.renderToString(
+                          CountryPopup(
+                            country, 
+                            state.language, 
+                            country.state?.studyConfig ? (() => history.push(`/${state.language}/${country.state.studyConfig.viewUrl}`)) : null )))
```
atatmaja, line 156: We seem to be mixing "undefined" and "null" as nullable states.

To prevent any weird errors I think we should either:
a) Decide on one of those nullable states and stick to it
b) be less specific when performing null checks (e.g. `if (map?.getSource(COUNTRY_LAYER_ID))` rather than `if (map?.getSource(COUNTRY_LAYER_ID) === undefined)`)

Option a is probably safer

src/components/map/Layers/StudyPins.tsx
```diff
@@ -89,8 +89,12 @@ const StudyPins = (map: mapboxgl.Map | undefined, records: AirtableRecord[]) =>
             pinPopup = new mapboxgl.Popup({ offset: 5, className: "pin-popup" })
               .setLngLat(e.lngLat)
               .setHTML(ReactDOMServer.renderToString(StudyPopup(record)))
-              .setMaxWidth("300px")
```
atatmaja, line 92: Does this do anything to styling on mobile?

src/components/map/Layers/StudyPins.tsx
```diff
@@ -99,16 +103,6 @@ const StudyPins = (map: mapboxgl.Map | undefined, records: AirtableRecord[]) =>
         });
       });
 
-      /*
```
atatmaja, line 102: Any reason why we have this commented vs deleted

src/components/map/Layers/Countries.tsx
```diff
@@ -1,50 +1,85 @@
...
+
+    map.removeFeatureState({
+        source: COUNTRY_LAYER_ID,
+        sourceLayer: COUNTRY_LAYER_ID
+        });
+
     estimateGradePrevalences.forEach((country: EstimateGradePrevalence) => {
         if (country && country.testsAdministered && country.alpha3Code) {
+
+            let partnershipConfig: any = PartnershipsConfig.Country.find(x => x.iso3 === country.alpha3Code)
```
annoyingpr, line 23: partnershipConfig should be typed here too based on Austin's comment from above

src/components/map/Popups/CountryPopup.tsx
```diff
@@ -32,13 +33,21 @@ const CountryPopup = (country : any, language : LanguageType) => {
...
           <div className="fit popup-text">{Translate("TestsAdministered")}: <b>{properties?.testsAdministered}</b></div>
           <div className="fit popup-text">{Translate('NumSeroprevalenceEstimates')}: <b>{properties?.numberOfStudies}</b></div>
         </div>
         {regions.map((o) => createPopupGeographySection(o.Region, o.Name))}
+        {onDetailsClick ? <Button 
+          className="not-found-button mt-2"
+          onClick={()=>{
+            onDetailsClick()
+          }}>
+          View Details
```
atatmaja, line 48: Translation needed. Also maybe this should be called "View Country Page"

src/components/map/Layers/Countries.tsx
annoyingpr: Nested ifs should be combined if possible

src/components/pages/Studies/Studies.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    height: "2800px",
+                  }}
+                />
+              </div>
+            </div>
+          </div>
+          <MaintenanceModal isOpen={isMaintenanceMode} headerText={""} />
+        </>
+      ) : (
+        <Redirect to="/ " />
```
atatmaja, line 87: Should we redirect to the 404 page to be more explicit about the fact that the partnerships page that the user is looking for doesn't exist?

src/components/map/Layers/Countries.tsx
```diff
@@ -54,80 +89,107 @@ function FilterCountryEstimates(map: mapboxgl.Map, estimateGradePrevalences: Est
...
-
-            map.on("mouseenter", "Countries", function (e: any) {
-                if (e.features[0].state.hasData) {
-                    countryPop.setMaxWidth("250px")
-                        .setHTML(ReactDOMServer.renderToString(CountryPopup(e.features[0], state.language)))
-                        .setLngLat(e.lngLat)
-                        //.trackPointer()
-                        .addTo(map);
-                }
+            const countryPop = new mapboxgl.Popup({
```
annoyingpr, line 137: maybe `countryPopup` is a better name

src/utils/MappingUtil.tsx
```diff
@@ -59,3 +59,41 @@ export function addEsriLayersFromVectorSourceStyle(style: any, map: mapboxgl.Map
...
+    xMin: undefined,
+    xMax: undefined,
+    yMin: undefined,
+    yMax: undefined,
+  }
+
+  coordinates.forEach((c)=>{
+    const longitude = c[0];
+    const latitude = c[1];
+    bbox.xMin = bbox.xMin < longitude ? bbox.xMin : longitude;
```
atatmaja, line 74: Nit: could use Math.max and Math.min instead

atatmaja (commented):
Good work so far, a few comments, ping me if any are unclear!

src/components/map/MapConfig.ts
```diff
@@ -22,6 +22,9 @@ export const MapSymbology = {
         }
     },
     CountryFeature: {
+        IsHighlighted: {
+            Color: '#b0c8d4'
```
annoyingpr, line 26: we should add specific colours as variables with SCSS so we don't have to hardcode them in code

src/components/map/MapConfig.ts
```diff
@@ -40,12 +43,21 @@ export const MapSymbology = {
...
 }
 
-export const MapUrlResource = {
+export const MapResources = {
     WHO_BASEMAP : "https://tiles.arcgis.com/tiles/5T5nSi527N4F7luB/arcgis/rest/services/WHO_Polygon_Basemap_no_labels/VectorTileServer",
     WHO_COUNTRY_VECTORTILES : "https://tiles.arcgis.com/tiles/5T5nSi527N4F7luB/arcgis/rest/services/Countries/VectorTileServer"
 }
 
-const MapConfig = {
+export const DefaultMapboxMapOptions = {
```
annoyingpr, line 51: we should type this!

src/components/map/MapboxMap.tsx
```diff
@@ -42,25 +43,32 @@ function mapOnLoad(map: mapboxgl.Map, dispatch: Dispatch<any>) {
...
-      const m = new mapboxgl.Map({
-        //@ts-ignore
-        container: mapContainerRef.current,
-        style: baseMapStyle,
-        center: [10, 30],
-        zoom: 2,
-        attributionControl: false,
-        antialias: true// enables MSAA antialiasing, to help make dotted line WHO borders more visible
-      });
+      const options = { // Merges options together to configure map
```
annoyingpr, line 62: more descriptive variable name like `mergedMapOptions` could replace this comment

src/components/shared/PartnershipsDropDown.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default function PartnershipsDropDown() {
+
+  const dropDownItemsRef = useRef([
+    // Add more items to this array if we have more partnerships
+    generateDropdownItem(PartnershipsConfig.Country[0])
+  ])
+
+
+  return (
+    <Dropdown text={Translate("Partnerships")} pointing className=".dropdown">
```
annoyingpr, line 17: can this work?
```suggestion
    <Dropdown text={Translate("Partnerships")} pointing className="dropdown">
```

src/httpClient.ts
```diff
@@ -171,6 +171,25 @@ export default class httpClient {
...
+            filters: filters,
+            include_subgeography_estimates: true,
+        }
+
+        const response = await this.httpPost('/data_provider/records', reqBody)
+        if (!response) {
+            return [];
+        }
+
+        const filtered_records = response.map((item: Record<string, any>) => {
```
annoyingpr, line 186: nit: we could use `forEach` here instead of `map` since we just want to mutate the original array right?

src/types.ts
```diff
@@ -57,6 +57,10 @@ export type Filters = {
   estimate_grade: any;
 };
 
+export type StudiesFilters = {
+  country: any;
```
annoyingpr, line 61: do we have a better type for this?

annoyingpr (changes requested):
awesome work - just a few nits and comments!

src/components/map/Layers/Countries.tsx
brett-stephen: yea there is a problem with the map loading to fast once its been cached. I need to look into refactoring this part without breaking it

src/utils/MappingUtil.tsx
brett-stephen: image

brett-stephen (commented):
made changes based on feedback,fixed layer loading issue

src/App.tsx
```diff
@@ -127,19 +128,8 @@ function App() {
...
-              width: "80vw",
-              height: "4100px"
-            }}
-            mobileOptions={{
-              width: "90vw",
-              height: "2800px"
-            }}
-          />
+        <Route path="/:language/Partnerships/:name">
+          <Studies />
```
atatmaja, line 132: nit: would it be a good idea to rename this component to match the page name?

brett-stephen: yea, will do 👍

src/components/map/Layers/Countries.tsx
```diff
@@ -1,133 +1,156 @@
...
-        map.on('sourcedata', function (e: any) {
-            if (e.sourceId === 'Countries' && map.isSourceLoaded('Countries')) {
-                map?.setFilter('Countries',
-                    ["in",
-                        ["get", 'CODE'],
-                        ["literal", onlyIso3]
-                    ]);
-            }
+function SetStudiesMode(map: mapboxgl.Map, iso3Name: string) {
+    map.removeFeatureState({
```
atatmaja, line 49: So it looks like we are completely wiping the map state here. I think this should work for the current use case (we only use map state to track isHighlighted, and only one country should be highlighted at a time), however if we end up using state to track other properties in the future this implementation wouldn't work as those would get wiped.

Can we add a comment describing this shortcoming to our code for future reference?

brett-stephen: yep, or I can just remove map highlight instead of doing a wipe. I'll look into how easy it would be to change

src/components/map/Layers/Countries.tsx
```diff
@@ -1,133 +1,156 @@
...
-        else if (map) {
-            map.on('sourcedata', function (e: any) {
-                if (e.sourceId === 'Countries' && map.isSourceLoaded('Countries')) {
+
+        if (map)
+        {
+            if(estimateGradePrevalences)
+            {
+                const sourceLoaded = map.getSource(COUNTRY_LAYER_ID) !== undefined
+                if (!sourceLoaded)
```
annoyingpr, line 80: are we okay with `map.getSource(COUNTRY_LAYER_ID) == null` in this case? if not, i think we can just do a null | undefined check this way 
```suggestion
                if (!map.getSource(COUNTRY_LAYER_ID) 
```

src/components/map/Layers/Countries.tsx
```diff
@@ -1,133 +1,156 @@
...
-
-            map.on("mouseenter", "Countries", function (e: any) {
-                if (e.features[0].state.hasData) {
-                    countryPop.setMaxWidth("250px")
-                        .setHTML(ReactDOMServer.renderToString(CountryPopup(e.features[0], state.language)))
-                        .setLngLat(e.lngLat)
-                        //.trackPointer()
-                        .addTo(map);
-                }
+            const countryPopup = new mapboxgl.Popup({
```
annoyingpr, line 99: can this be typed?

atatmaja (approved):
A few pretty minor comments, LGTM!

annoyingpr (approved):

src/components/map/Layers/Countries.tsx
brett-stephen: Just left a comment for now.

src/components/map/Layers/Countries.tsx
brett-stephen: I believe its implicitly typed from mapbox's own popup type

src/components/pages/Partnerships/Partnerships.tsx
```diff
@@ -47,7 +49,21 @@ export default function Partnerships() {
                   })}
                 </h3>
               </div>
-              <div style={{ height: "70vh" }} className="row">
+              { isMobileDeviceOrTablet ?
```
atatmaja, line 52: Nit: this looks a bit messy, my preference would be to return both the mobile/desktop views here from a function so that the jsx ternary looks cleaner here

src/utils/translate/en.json
```diff
@@ -363,7 +363,11 @@
     "Canada": "Canada - Covid-19 Immunity Task Force"
   },
   "PartnershipsPageDescription": {
-    "Canada": "SeroTracker is financially supported by the Public Health Agency of Canada through the COVID-19 Immunity Task Force (CITF). The CITF has funded numerous high calibre serosurveys amongst other research which you can learn about here. This page showcases the data from Canadian serosurveys, and modeling based on these data."
+    "Canada": "SeroTracker is financially supported by the Public Health Agency of Canada through the COVID-19 Immunity Working Group (CITF). CITF has funded many high-calibre serological surveys among other research that you can learn more about on their website. This page presents data from Canadian serological surveys and the modelling based on these data."
+  },
+  "PartnershipsView": {
+    "MapNote1": "The map displays estimates for all geographic based estimates. Pins are arranged around the center of their geographic location(s).",
+    "MapNote2": "Case and death counts (at the midpoint in study sampling) are taken at a national level from WEBSITE (subsource: coronavirus.jhu.edu)"
```
atatmaja, line 373: Does "subsource" turn into a hyperlink when rendered?? Wording looks a bit weird if otherwise

src/components/map/MapboxMap.tsx
```diff
@@ -68,7 +68,7 @@ const MapboxMap = ( {mapConfig, countriesConfig, studyPinsConfig}: MapboxMapProp
         ...mapConfig,
       };
 
-      const m = new mapboxgl.Map(mergedOptions);
+      const m = new mapboxgl.Map(mergedOptions).addControl(new mapboxgl.NavigationControl());
```
atatmaja, line 71: This just adds nav controls to the map right?

src/components/map/Layers/Countries.tsx
```diff
@@ -45,31 +45,8 @@ function SetCountryEstimates(map: mapboxgl.Map, estimateGradePrevalences: Estima
     });
 }
 
-function SetMapData(map: mapboxgl.Map, estimateGradePrevalences: EstimateGradePrevalence[], countryFocus: any) {
-
+function SetMapData(map: mapboxgl.Map, estimateGradePrevalences: EstimateGradePrevalence[]) {
     SetCountryEstimates(map, estimateGradePrevalences);
-
-    if(countryFocus) // Shades following countries dark
```
atatmaja: any reason why we are getting rid of this countryFocus behaviour? don't think it's essential but just want to be on the same page

atatmaja:
LGTM still, a few questions and nit requests that might be worth to consider but giving you the green light @brett-stephen

src/components/map/MapboxMap.tsx
brett-stephen: yep

src/components/map/Layers/Countries.tsx
brett-stephen: found a way to highlight the country without actually needing this mode

brett-stephen: We just limit the sero data shading to only one country, effecting making the countryfocus mode obsolete

src/utils/translate/en.json
brett-stephen: I don't think it will be a link. Will need to fix that

src/components/pages/Partnerships/Partnerships.tsx
brett-stephen: you're right, just wanted to fix the view on mobile so that it wasn't completely broken

brett-stephen: I'll add a TODO

PR #248, opened 2021-06-10
Unity studies filter

atatmaja:

- Adds unity studies checkbox and /Unity view (same as explore but with checkbox checked off by default)
- Changes filters and filter options data models to use arrays instead of sets (sets were an artifact from when we were doing filtering logic on the frontend)
- Removes some unnecessary code (dataPageState, some types that were not being used, API call that's not being used, unnecessary dispatches)

https://airtable.com/tbli2lWQHAqBa6ZcI/viwWvxb27XGuziw4c/recGnymO05NXQZDqX?blocks=bipIkdtYHmboUfJut

image
image
image
image

Smoke test

Yes, but PR has been merged

Yes, will include them all once we figure out exactly what text we want to show

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -156,7 +156,7 @@ function App() {
         </Route>
         <Route path="/:language/Unity">
           <Explore initialFilters={(() => {
-            const initialFilters = getEmptyFilters();
+            let initialFilters = getEmptyFilters();
             initialFilters.unity_aligned_only = true;
             return initialFilters;
           })()}/>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,10 +13,9 @@ import { CookieBanner } from "./components/shared/CookieBanner";
 import { NavBar } from "./components/shared/NavBar";
 import { Footer } from "./components/shared/Footer";
 import TableauEmbed from "./components/shared/TableauEmbed";
-import { AppContext } from "./context";
+import { AppContext, getEmptyFilters } from "./context";
 import httpClient from "./httpClient";
 import { LanguageType, PageStateEnum } from "./types";
-import { initializeData } from "./utils/stateUpdateUtils";
 import { setLanguageType } from "./utils/translate/translateService";
 import { ANALYZE_URLS, CANADA_URLS } from "./constants";
 
@@ -26,8 +25,7 @@ function App() {
 
   // General call that happens once at the start of everything.
   useEffect(() => {
-    const api = new httpClient()
-    initializeData(dispatch, explore.filters, PageStateEnum.explore)
+    const api = new httpClient();
     const allFilterOptions = async () => {
       const { options, updatedAt, maxDate, minDate } = await api.getAllFilterOptions();
       dispatch({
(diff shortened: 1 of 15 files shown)


src/components/shared/InformationIcon.tsx
```diff
@@ -23,6 +23,7 @@ export default function InformationIcon(props: InformationIconProps) {
         offset={offset}
         position={position}
         size={popupSize}
+        hoverable={true}
```
annoyingpr, line 26: nit: this should work too right?
```suggestion
        hoverable

```

src/utils/translate/en.json
```diff
@@ -448,8 +449,17 @@
...
   "StudyInformation": "Study Information",
   "StudyInformationTooltip": {
-    "FirstParagraph": "Filter on study details, including source type (publication, preprint, news, or report) and study status (ongoing, completed).",
-    "SecondParagraph": "Risk of bias: Reflects the extent to which the true prevalence may be different from the estimated prevalence. Estimated by SeroTracker reviewers based on the Joanna Briggs Institute critical appraisal tool for prevalence estimates."
+    "FirstParagraph": "Filter on study details, including source type (publication, preprint, news, or report), study status (ongoing, completed), and alignment with the WHO Unity protocol.",
+    "SecondParagraph": {
+      "PartOne": "Risk of bias: ",
+      "PartTwo": "Reflects the extent to which the true prevalence may be different from the estimated prevalence. Estimated by SeroTracker reviewers based on the Joanna Briggs Institute critical appraisal tool for prevalence estimates."
+    },
+    "ThirdParagraph": {
```
annoyingpr, line 457: I know we're overhauling translation soon, but think we should not include extra spaces at the end of text in this JSON, that should be handled elsewhere.

src/components/sidebar/right-sidebar/Filters.tsx
```diff
@@ -95,12 +83,26 @@ export default function Filters({ page }: FilterProps) {
...
             })
           }}
-          defaultValue={getFilters(filter_type)}
+          defaultValue={pageState.filters[filter_type] as string[]}
         />
       </div>
     )
   }
 
+  const buildFilterCheckbox = (filter_type: FilterType, label: string) => {
```
annoyingpr, line 92: Not sure how this worked before, but what is this used for & why was this added?

src/App.tsx
```diff
@@ -156,6 +154,13 @@ function App() {
         <Route path="/:language/Insights">
           <Insights />
         </Route>
+        <Route path="/:language/Unity">
+          <Explore initialFilters={(() => {
+            const initialFilters = getEmptyFilters();
+            initialFilters.unity_aligned_only = true;
```
annoyingpr, line 160: Didn't know you could update `const` - but read more about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const. JavaScript is weird

annoyingpr (approved):
Just a few questions but LGTM 🚀

src/App.tsx
atatmaja: yeah since we are not technically updating the reference we can update the const like this. this is a bit sketchy tho, I will change it to a `let` so what we're doing here is more apparent

src/components/sidebar/right-sidebar/Filters.tsx
atatmaja: This is a new function that renders a checkbox filter rather than our typical dropdown filter. Needed to create this for the unity studies checkbox that was newly implemented

PR #254, opened 2021-07-07
Bug fix - all seroprevalence estimates showing as NULL on study modal

simonarocco:
- Seroprevalence field was showing as N/A for all study modals
- This was because the backend field is called serum_pos_prevalence but the frontend was calling record.seroprevalence

- Looked at study modals with and without different filters


atatmaja (approved):
LGTM as long as we've tested thoroughly

PR #260, opened 2021-07-31
Himanshu/date picker

HimanshuRanka:

**Bugs**: Year dropdown was not working. No slider for picking dates

https://airtable.com/tbli2lWQHAqBa6ZcI/viw2Kjofxz0J1dHD0/recCIck2AZQs29GNC?blocks=bipIkdtYHmboUfJut

Old:
image
image

New: 
image
image

Manually tested different dates and slider positions and looked if it was updating the map correctly. 

N/A

N/A

### Explore

- Check Desktop and on Mobile
- Summary Stats
- Map has pins
- Filter info bubbles

### Analyse

### Data
- Click each button
    - Download the CSV
- "terms of use" button

### Publications
- Cards all load with images
- Carousel is clickable

### About
- Page loads
- Scroll to bottom

- Check in French

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
--- a/src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
+++ b/src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
@@ -2,8 +2,8 @@ import React, {useCallback, useEffect, useRef, useState} from 'react';
 import './DateSlider.css';
 
 interface DSprops {
-    min: number; //minimum possible value for slider
-    max: number; //maximum possible value for slider
+    minPossibleValue: number; //minimum possible value for slider
+    maxPossibleValue: number; //maximum possible value for slider
     values: Date[];
     minDate: Date;
     onChange: Function;
@@ -12,20 +12,20 @@ interface DSprops {
 export default function DateSlider(props: DSprops){
 
     const milliSecondsPerDay = 86400000;
-    const {min, max, minDate} = props;
+    const {minPossibleValue, maxPossibleValue, minDate} = props;
     const [sliderThumbValues, setSliderThumbValues] = useState<number[]>([(props.values[0].getTime() - minDate.getTime())/milliSecondsPerDay,
         (props.values[1].getTime() - minDate.getTime())/milliSecondsPerDay])
     const [leftThumbVal, setLeftThumbVal] = useState(sliderThumbValues[0]);
     const [rightThumbVal, setRightThumbVal] = useState(sliderThumbValues[1]);
-    const minValRef = useRef(min);
-    const maxValRef = useRef(max);
+    const minValRef = useRef(minPossibleValue);
+    const maxValRef = useRef(maxPossibleValue);
     const range = useRef<HTMLDivElement>(null);
     /*const getPercent = useCallback(
         (value: number) => Math.round(((value - min) / (max - min)) * 100),
         [min, max]
     );*/
 
-    const getPercent = (value: number) => Math.round(((value - min) / (max - min)) * 100)
+    const getPercent = (value: number) => Math.round(((value - minPossibleValue) / (maxPossibleValue - minPossibleValue)) * 100)
 
     useEffect(() => {
       setLeftThumbVal(sliderThumbValues[0]);
@@ -64,7 +64,7 @@ export default function DateSlider(props: DSprops){
 
     return(
         <div className={"slider-container"}>
-            <input className={"thumb thumb-left"} type={"range"} value={leftThumbVal} min={min} max={max}
+            <input className={"thumb thumb-left"} type={"range"} value={leftThumbVal} min={minPossibleValue} max={maxPossibleValue}
                    onChange={event => {
                        const value = Math.min(Number(event.target.value), rightThumbVal - 1);
                        setLeftThumbVal(value);
@@ -74,9 +74,9 @@ export default function DateSlider(props: DSprops){
                        props.onChange(true, toDateSinceMinDate(leftThumbVal));
                    }}
 
-                   style={{ zIndex: (leftThumbVal > max - 100 ? 5 : 3)}}
+                   style={{ zIndex: (leftThumbVal > maxPossibleValue - 100 ? 5 : 3)}}
             />
-            <input className={"thumb thumb-right"} type={"range"} value={rightThumbVal} min={min} max={max}
+            <input className={"thumb thumb-right"} type={"range"} value={rightThumbVal} min={minPossibleValue} max={maxPossibleValue}
                    onChange={event => {
                        const value = Math.max(Number(event.target.value), leftThumbVal + 1);
                        setRightThumbVal(value);
--- a/src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
+++ b/src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
@@ -6,24 +6,20 @@ interface DSprops {
     maxPossibleValue: number; //maximum possible value for slider
     values: Date[];
     minDate: Date;
-    onChange: Function;
+    onMouseUp: Function;
 }
 
-export default function DateSlider(props: DSprops){
+export default function DateSlider({minPossibleValue, maxPossibleValue, minDate, values, onMouseUp}: DSprops){
 
     const milliSecondsPerDay = 86400000;
-    const {minPossibleValue, maxPossibleValue, minDate} = props;
-    const [sliderThumbValues, setSliderThumbValues] = useState<number[]>([(props.values[0].getTime() - minDate.getTime())/milliSecondsPerDay,
-        (props.values[1].getTime() - minDate.getTime())/milliSecondsPerDay])
+
+    const [sliderThumbValues, setSliderThumbValues] = useState<number[]>([(values[0].getTime() - minDate.getTime())/milliSecondsPerDay,
+        (values[1].getTime() - minDate.getTime())/milliSecondsPerDay])
     const [leftThumbVal, setLeftThumbVal] = useState(sliderThumbValues[0]);
     const [rightThumbVal, setRightThumbVal] = useState(sliderThumbValues[1]);
     const minValRef = useRef(minPossibleValue);
     const maxValRef = useRef(maxPossibleValue);
     const range = useRef<HTMLDivElement>(null);
-    /*const getPercent = useCallback(
-        (value: number) => Math.round(((value - min) / (max - min)) * 100),
-        [min, max]
-    );*/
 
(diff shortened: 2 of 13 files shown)


.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
atatmaja, line 1: let's not commit from this .idea folder and add it to our gitignore

src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+    max: number;
+    values: Date[];
+    minDate: Date;
+    onChange: Function;
+}
+
+export default function DateSlider(props: DSprops){
+
+    const {min, max, minDate} = props;
+    const [values, setValues] = useState<number[]>([(props.values[0].getTime() - minDate.getTime())/86400000,
```
atatmaja, line 15: is there a better way to do this? if not can we at least assign 86400000 to a variable with a descriptive name (say secondsPerYear or something) for better readability?

src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+    useEffect(() => {
+      setMinVal(values[0]);
+      setMaxVal(values[1]);
+    }, [values])
+
+    useEffect(() => {
+        const minPercent = getPercent(minVal);
+        const maxPercent = getPercent(maxVal);
+        console.log(minPercent, maxPercent);
```
atatmaja, line 36: remove console log

src/httpClient.ts
```diff
@@ -69,14 +69,29 @@ export default class httpClient {
...
+                options[k] = new Set(response[k]);
+            }
+        }
         // We know that only these 3 isotypes will ever be reported, thus we can hardcode
         options.isotypes_reported = ["IgG", "IgA", "IgM"];
         const updatedAt = format(parseISO(response.updated_at), "yyyy/MM/dd");
-        const maxDate = parseISO(response.max_date);     
-        const minDate = parseISO(response.min_date);
+
+        const maxDate = parseISO(response.max_publication_end_date);
```
atatmaja, line 90: was this a change made recently??

src/httpClient.ts
```diff
@@ -69,14 +69,29 @@ export default class httpClient {
...
     async getAllFilterOptions() {
         const response = await this.httpGet('/data_provider/filter_options', true);
-        const options = {...response} as Record<string, any>;
+
+        const options: Record<string, any> = {}
+        for(let k in response){
+            // Currently no need for max and min date options
+            //Note: From Himanshu: I changed the values of the string here
+            // because I noticed the backend response did not have a max_date and min_date value
+            //but it did have what is below.
```
atatmaja, line 78: Remind me to check this in the backend code tomorrow, good spot!

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
```diff
@@ -59,53 +81,70 @@ export default function Datepicker({ page }: DatepickerProps) {
...
+                onSelect={(date: Date) => datePickerChanged(false, date)}
+                dateFormat="yyyy/MM/dd"
+                dateFormatCalendar={"MMM yyyy"}
+                minDate={startDate}
+                maxDate={maxDate}
+                showYearDropdown
+                yearDropdownItemNumber={5}
+                scrollableYearDropdown
+                showMonthDropdown
+
```
atatmaja, line 136: nit remove empty line

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
```diff
@@ -59,53 +81,70 @@ export default function Datepicker({ page }: DatepickerProps) {
     </div>
   )
 
+  const handleSliderChange = (min: number, max: number) => {
```
atatmaja, line 84: delete unused function declaration

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
```diff
@@ -16,15 +21,32 @@ interface DatepickerProps {
...
-  const [filterStartDate, filterEndDate] = pageState.filters.publish_date;
-  const [startDate, setStartDate] = useState(filterStartDate);
-  const [endDate, setEndDate] = useState(filterEndDate);
+  const [filterStartDate, filterEndDate]: Date[] = Array.from(pageState.filters.publish_date);
+  const [startDate, setStartDate] = useState<Date>(filterStartDate);
+  const [endDate, setEndDate] = useState<Date>(filterEndDate);
+  const minDate = state.calendarStartDates.minDate;
+  const maxDate = state.calendarStartDates.maxDate;
+  const [sliderMax, setSliderMax] = useState<number>((maxDate.getTime()-minDate.getTime())/86400000);
+  const [values, setValues] = useState<Date[]>([startDate, endDate]);
```
atatmaja, line 30: Can we be a bit clearer with these state variable names to differentiate state variables that pertain to the datepicker vs the slider

atatmaja:
Are the datepicker and slider perfectly in sync (i.e. if I change the dates on the slider, will the datepicker UI be updated to reflect the changes and vice versa)? I wasn't completely sure since the two components seem to be rendered using different state variables. 

From a usability perspective I think this would be necessary to not confuse our users

src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const {min, max, minDate} = props;
+    const [values, setValues] = useState<number[]>([(props.values[0].getTime() - minDate.getTime())/86400000,
+        (props.values[1].getTime() - minDate.getTime())/86400000])
+    const [minVal, setMinVal] = useState(values[0]);
+    const [maxVal, setMaxVal] = useState(values[1]);
+    const minValRef = useRef(min);
+    const maxValRef = useRef(max);
+    const range = useRef<HTMLDivElement>(null);
+
+    const getPercent = useCallback(
```
atatmaja, line 23: just curious & trying to learn more, why are we using useCallback here???

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
```diff
@@ -16,15 +21,32 @@ interface DatepickerProps {
...
   useEffect(() => {
     registerLocale("en", enUS)
     registerLocale("fr", fr)
   }, [])
 
+  useEffect(() => {
+    setValues([startDate, endDate]);
+  }, [startDate, endDate])
+
+  useEffect(() => {
```
atatmaja, line 45: Is this useffect really necessary? If there is no way to access dates outside the min/max range in the UI, don't think guarding against this edge case is really a practical concern. 

Also if we were to guard against this edge case, we would probably want to ensure that out of range dates aren't even stored in the context (pageState.filters.publish_date) in the first place.

src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
HimanshuRanka: useCallBack makes sure that the function is only redefined when one of the deps changes, instead of redefining it every rerender. So getPercent would be redefined if min or max changed (which are passed in as props). which would then trigger the useEffect causing the min and max percent to change moving the rail bar to the right position. by using useCallBack we cause a rerender or recalculation if and only if the min and max values of the whole bar change which is highly unlikely (possible but rare). and so this improves performance by reducing rerendering. 

That being said, I only really leart what useCallBack does while going through the tutorial for the slider and then looked it up a little more to understand, and deduced the above reasoning which makes sense to me. SO if you feel like I might be misunderstanding its use or that the performance gain really is not that great, I am open to criticism.

HimanshuRanka: Also all that being said, the more I am a little more confused as to whether it is needed here or not. @brett-stephen @annoyingpr if you guys have any input on this, it is more than welcome as well.

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
HimanshuRanka: Yeah you are right!

HimanshuRanka:
> users

The date picker and slider are almost perfectly insync. Just to prevent a continuous flurry of calls to the backend, I made it so that the slider only updates the input field and actual date onMouseUp instead of onChange of the slider itself. That way while a user is dragging the slider it does not change the value and does not bombard the backend with requests. As soon as the values are picked the date is changed

src/components/sidebar/right-sidebar/datepicker/Datepicker.tsx
HimanshuRanka: Actually I noticed that although it isnt possible for the user to cause the above bug. The server loading speed does cause it initially which can cause the following 2 issues on first try:

image

Where as you can see the dates on the slider are updated to the correct min and max, however the datepicker boxes have the old dates that are out of range. This causes another problem when you move the slider:

image

where it then sets the min date to the date that is out of range, and the blue bar runs off screen and detatches from the thumbs.

I changed the code a little so that the above check is only done when the server min and max are changed, but I dont think there is another way to get around this

HimanshuRanka: Fixed in b5cfb7a067d187dc9bcd1aef3fd4edfb98a416f0

src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+    const milliSecondsPerDay = 86400000;
+    const {minPossibleValue, maxPossibleValue, minDate} = props;
+    const [sliderThumbValues, setSliderThumbValues] = useState<number[]>([(props.values[0].getTime() - minDate.getTime())/milliSecondsPerDay,
+        (props.values[1].getTime() - minDate.getTime())/milliSecondsPerDay])
+    const [leftThumbVal, setLeftThumbVal] = useState(sliderThumbValues[0]);
+    const [rightThumbVal, setRightThumbVal] = useState(sliderThumbValues[1]);
+    const minValRef = useRef(minPossibleValue);
+    const maxValRef = useRef(maxPossibleValue);
+    const range = useRef<HTMLDivElement>(null);
+    /*const getPercent = useCallback(
```
brett-stephen, line 23: can we remove the commented code?

src/components/sidebar/right-sidebar/datepicker/DateSlider.tsx
```diff
@@ -0,0 +1,97 @@
...
+    maxPossibleValue: number; //maximum possible value for slider
+    values: Date[];
+    minDate: Date;
+    onChange: Function;
+}
+
+export default function DateSlider(props: DSprops){
+
+    const milliSecondsPerDay = 86400000;
+    const {minPossibleValue, maxPossibleValue, minDate} = props;
```
brett-stephen, line 15: Nit: It might be cleaner to destructure the props within the parameter on line 12 like `DateSlider({minPossibleValue, maxPossibleValue, minDate}: DSprops)`

brett-stephen (approved):
couple small things, but otherwise looks good

PR #281, opened 2021-09-18
Footer fixes

brett-stephen:
- fixed filters sidebar issues,
- removed ununsed code,
- fixed white link color out of footer

- general issue clean up from the footer update


src/components/sidebar/right-sidebar/RightSidebar.tsx
```diff
@@ -1,8 +1,6 @@
 import React from "react";
 import "../sidebar.scss";
-import Datepicker from "./datepicker/Datepicker";
```
HimanshuRanka, line 3: is the datepicker not a part of the right sidebar?

src/types.ts
```diff
@@ -42,7 +42,7 @@ export type AggregatedRecord = {
   seroprevalence: number;
 };
 
-export type Filters = {
+export type FiltersConfig = {
```
HimanshuRanka, line 45: why the name change?

HimanshuRanka (approved):
LGTM! just a couple random questions

src/types.ts
brett-stephen: We already had a `Filters` component, so a `Filters` type was just confusing and hard to work with

src/components/sidebar/right-sidebar/RightSidebar.tsx
brett-stephen: It is, but I believe this was unused code from a different implementation

HimanshuRanka (approved):

PR #295, opened 2021-11-04
New Github download button

serosurveillance-can:
### Explore

- Check Desktop and on Mobile
- Summary Stats
- Map has pins
- Filter info bubbles

### Analyse

### Data
- Click each button
    - Download the CSV
- "terms of use" button

### Publications
- Cards all load with images
- Carousel is clickable

### About
- Page loads
- Scroll to bottom

- Check in French


src/components/shared/DownloadButton.tsx
```diff
@@ -42,22 +43,40 @@ export default function DownloadButton(props: DownloadButtonProps) {
     }
     
     return (
-        <Button key={buttonLabelKey} size={buttonSize} className="download-data-btn">
+        iconName === "" ?
```
atatmaja, line 46: Since this needs to be done on a time crunch I won't let this block merge. But I don't think this is an extensible or clear design - nobody will understand that iconName == "" translates to the airtableLogo.

Moving forward I think we should let the user specify some jsx that would be rendered as the "iconComponent", or EITHER an iconName or iconSource.

src/components/shared/DownloadButton.tsx
```diff
@@ -42,22 +43,40 @@ export default function DownloadButton(props: DownloadButtonProps) {
     }
     
     return (
-        <Button key={buttonLabelKey} size={buttonSize} className="download-data-btn">
+        iconName === "" ?       
+        <Popup content={Translate(popupText)} trigger={
+            <Button key={buttonLabelKey} size="large" iconName={iconName}className="download-data-btn mb-2 mr-2">
```
atatmaja, line 48: Again not a big deal for now but we should be using the buttonSize prop here.

Also I see that we've defined some hardcoded width and heights, ideally we should have some mapping from iconSize to height/width and allow users to specify iconSize. That's more of a nitpick tho

src/components/shared/DownloadButton.tsx
atatmaja: @annoyingpr can we make this a flup ticket

atatmaja (approved):
Not leaving a blocking review because of time crunch, but we need to do some cleanup.

PR #323, opened 2021-12-22
German Translation

annoyingpr:
German translation for RKI

[Implement German localization](https://airtable.com/appT8tnLbGJV6v81V/tbli2lWQHAqBa6ZcI/viwTYqtBDdt3Wt3Yo/recx2CsTZebPiqXu4?blocks=hide)

### Explore

- Check Desktop and on Mobile
- Summary Stats
- Map has pins
- Filter info bubbles

### Analyse

### Data
- Click each button
    - Download the CSV
- "terms of use" button

### Publications
- Cards all load with images
- Carousel is clickable

### About
- Page loads
- Scroll to bottom

- Check in French


src/components/pages/Publications/PublicationsCard.tsx
```diff
@@ -9,6 +9,19 @@ import {PublicationProps} from "./PublicationsConstants"
 
 export function PublicationsCard(props: PublicationProps) {
   const [state] = useContext(AppContext);
+
+  const dateFormatsByLanguage = {
```
atatmaja, line 13: Slightly cleaner pattern would be to make a shared utility function to perform this logic (you could use it both in the publicationsCard and publicationsItem components). It could have the following signature:

```
function getFormattedDate(day, month, year, language)
```

src/components/shared/Header.tsx
```diff
@@ -30,156 +33,197 @@ export const Header = () => {
...
 
-  const changeLanguages = () => {
+  const changeLanguages = (newLanguage: LanguageType) => {
     let url = history.location.pathname;
     if (state.language === LanguageType.english) {
-      url = url.replace("/en/", "/fr/")
-    }
-    else {
-      url = url.replace("/fr/", "/en/")
+      url = url.replace("/en/", `/${newLanguage}/`)
```
atatmaja, line 39: Could you not just do:

```
url = url.replace(`/${state.language}/`, `/${newLanguage}/`)
```

instead of using all these if statements

src/utils/translate/translateService.ts
```diff
@@ -64,8 +67,15 @@ export default function Translate(
   substitution: Record<string, string | number> | null = null,
   addSpaces: [boolean, boolean] | null = null
 ): string {
-  const translationDictionary: Json =
-    language === LanguageType.english ? (English as Json) : (French as Json);
+  let translationDictionary: Json = {}
+
+  if (language === LanguageType.english) {
```
atatmaja, line 72: Wouldn't it be easier to just make this a lookup object as discussed, something like:

```
const translations = {
  "en": English,
  "fr": French,
  "de": German
}

const translationDictionary: Json = translations[language]
```

src/utils/utils.tsx
```diff
@@ -79,3 +79,28 @@ export const NumberDateToWordDate = (date_str: string) => {
   const monthStr = months[Number(month) - 1];
   return `${Translate("Months", [monthStr])} ${day}, ${year}`
 }
+
+export const getformattedDate = (day: string | undefined, month: string | undefined, year: string | undefined, language: LanguageType) => {
```
atatmaja, line 83: might be simpler to use a switch statement here, something like:

`
switch(language){   
        case 'en': return `${month } ${day ? day + "," : ""} ${year}`;
        case 'fr': return `${day} ${Translate("Months", [month])} ${year}`;
        case 'de': return `${day ? day + "." : ""} ${Translate("Months", [month])} ${year}`;
        default: return `${month } ${day ? day + "," : ""} ${year}`;      
    }
`

src/utils/utils.tsx
```diff
@@ -79,3 +79,28 @@ export const NumberDateToWordDate = (date_str: string) => {
   const monthStr = months[Number(month) - 1];
   return `${Translate("Months", [monthStr])} ${day}, ${year}`
 }
+
+export const getformattedDate = (day: string | undefined, month: string | undefined, year: string | undefined, language: LanguageType) => {
+  interface dateFormatsByLanguageOptions {
```
atatmaja, line 84: Probably simpler to use a switch statement here, something like:

```suggestion
switch(language){   
        case 'en': return `${month } ${day ? day + "," : ""} ${year}`;
        case 'fr': return `${day} ${Translate("Months", [month])} ${year}`;
        case 'de': return `${day ? day + "." : ""} ${Translate("Months", [month])} ${year}`;
        default: return `${month } ${day ? day + "," : ""} ${year}`;      
}
```

src/components/shared/Header.tsx
```diff
@@ -30,156 +33,191 @@ export const Header = () => {
...
+        <Accordion.Title
+              active={languageDropdownActive}
+              index={0}
+              onClick={handleLanguageClick}
+              className={'flex py-0'}
+            >
+              {state.language.toUpperCase()}
+              <Icon name='dropdown' className={`ml-5 ${languageDropdownActive ? 'dropdown__icon' : 'dropdown__icon--active'} `}/>
+              </Accordion.Title>
+              { state.language !== LanguageType.english ?
```
HimanshuRanka, line 114: Confused as to why this is a !== LanguageType.english because wouldn't that mean 2/3 cases are always active (unless that is what is expected)?

HimanshuRanka: OHHH so if a certain language is active, it does not exist in the dropdown? Did I understand correct

src/components/shared/Header.tsx
```diff
@@ -30,156 +33,191 @@ export const Header = () => {
...
+        <Dropdown.Item onClick={() => {changeLanguages(LanguageType.french)}}>
+        { LanguageType.french.toUpperCase() }
+        </Dropdown.Item> : ""}
+     {state.language !== LanguageType.german ?      
+         <Dropdown.Item onClick={() => {changeLanguages(LanguageType.german)}}  >
+        { LanguageType.german.toUpperCase() }
+        </Dropdown.Item> : ""}
+    </Dropdown.Menu>)
+  
+  const renderPartnershipsAccordianContent = (partnershipDrownDownActive: boolean) => (
```
HimanshuRanka, line 203: What is the difference between this method and the renderParnershipDropdownMenu method?

HimanshuRanka: Is one for desktopnav and the other for mobilenav? Could there be a better way to do that instead of 2 very similar methods (accordian vs dropdown is the only difference right)?

HimanshuRanka (commented):
Just had one discussion point I wanted to address

src/components/shared/Header.tsx
annoyingpr: yeah exactly!

src/components/shared/Header.tsx
annoyingpr: i didn't set up these methods, but I think it makes sense to have them separate since the content is pretty different between the 2

atatmaja (approved):
LGTM

HimanshuRanka (approved):

PR #346, opened 2022-02-14
Lang diff script

HimanshuRanka:

added a couple scripts to help maintain the language files we use. 

https://airtable.com/appT8tnLbGJV6v81V/tbli2lWQHAqBa6ZcI/viw2Kjofxz0J1dHD0/rec8ZwsVfrRyfvrm6?blocks=hide

### Explore

- Check Desktop and on Mobile
- Summary Stats
- Map has pins
- Filter info bubbles

### Analyse

### Data
- Click each button
    - Download the CSV
- "terms of use" button

### Publications
- Cards all load with images
- Carousel is clickable

### About
- Page loads
- Scroll to bottom

- Check in French


src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
+#Translations
```
simonarocco, line 1: should be # Translations (add the space otherwise it doesn't render properly)

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
...
+
+How do they work:
+
+We have 3 files named en.json, fr.json and de.json that contain english, french, and german translations of all the content we display on the website.
```
simonarocco, line 5: change to `en.json`, `fr.json`, `de.json` so it renders a bit nicer

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
+#Translations
+
+How do they work:
+
+We have 3 files named en.json, fr.json and de.json that contain english, french, and german translations of all the content we display on the website. 
+
+We use a Translate() function inaccross the website whenever displaying content that renders the correct translation for every section.
```
simonarocco, line 7: inacross --> across

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
+#Translations
+
+How do they work:
+
+We have 3 files named en.json, fr.json and de.json that contain english, french, and german translations of all the content we display on the website. 
+
+We use a Translate() function inaccross the website whenever displaying content that renders the correct translation for every section. 
+
+the first argument of the function is the key. This can either map to a translation or dictionary of values.
```
simonarocco, line 9: The capital

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
...
+The second argument is an array of secondary keys that form a tree like structure. each key can either be a translation or another dictionary and is created in a nested manor
```
simonarocco, line 10: "Each" and "manner."

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Maintainance
+
+We have an inhouse team of individuals fluent in french that help us accurately translate from english to french.
+
+Whenever new content is added to the website, the english and french translations are posted on a translations board and approved/corrected by the team.
+
+The German translations are taken care of by the RKI. we have a checkpoint branch in git which we then run a diff on to get all the changes made
```
simonarocco, line 18: Just capitalize words at beginning of sentence

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
...
+```
+    get_diff.py
+    
+    cleanup_translate.py
+```
+
+get_diff.py runs a diff on all 3 json files and outputs which keys are not in sync
+i.e.
+- Missing from one of the files
+- Mispelled
```
simonarocco, line 33: Misspelled

src/utils/translate/Translations.md
```diff
@@ -0,0 +1,43 @@
...
+- Deprecated
+
+These are then printed out to the console and can be used to fix the issues.
+
+The cleanup_translate.py file on the other hand goes through every file in the codebase
+and comes up with a list of all the keys actually used by Translate functions. 
+This way any translations and clutter that exists in our files that we no longer use can be removed improving
+the readability of the translations. 
+
+Both these scripts are very bare bones and have a decent scope for imporvement, but that can come later
```
simonarocco, line 43: improvement

src/cleanup_translate.py
```diff
@@ -0,0 +1,26 @@
+import os
```
simonarocco, line 1: Can you comment this please? Even high level just high level what each for loop is doing. also can you give an example of the regex that you are searching for. also, let's put this in a function main() and call main() at the bottom of the script.

simonarocco: Also, wouldn't it be better to come up with a list of keys that aren't used? so you would get the list of keys that are used, look at the json files and see all the keys that are there, and then subtract the keys that are used to get the ones we don't use. if that makes sense. cause right no translate_keys.json contains a ton of keys because we use lots of them, but it doesn't highlight which ones we DONT use and those are the ones we acc care about and what to remove

src/utils/translate/get_diff.py
```diff
@@ -0,0 +1,40 @@
...
+de = json.load(open("de.json", "r", encoding="utf-8"))
+
+
+def get_diff(dict1, dict2):
+    missing_keys = []
+    for key in dict1:
+        if key not in dict2:
+            missing_keys.append(key)
+        elif type(dict1[key]) is dict:
+            keys = get_diff(dict1[key], dict2[key])
```
simonarocco, line 14: ham recursion. very impressive

src/utils/translate/get_diff.py
```diff
@@ -0,0 +1,40 @@
+import json
+
+en = json.load(open("en.json", "r", encoding="utf-8"))
```
simonarocco, line 3: let's just turn this into a function that accepts to file paths and returns the missing keys. much easier to reuse and run. to be honest, let's also make cleanup translate a function too and put these two functions in a python file called translate_utils and then we just have two functions that we can call when we need. that way you can call get_diff with old and new german file paths and you don't have stuff hardcoded so it's more extendable. does that make sense? then we just call these functions on the fly when we want to update de.json or clean up the keys in these files

simonarocco (changes requested):
see my comments

PR #340, opened 2022-02-06
Update README.md to include more details surrounding node setup

atatmaja:
### Explore

- Check Desktop and on Mobile
- Summary Stats
- Map loads in ~8 seconds with loading spinner 
- Map has pins
- Country Modal - check all fields
- Pins modal (check each grade) - check all fields
- Filter info bubbles
- Select each filter and check options
- Execute a Filter for Risk of bias, Sex, and Date.
- Check the "Database up to date to:" date
- Check each footer item (Privacy policy etc.)

### Analyse
- Check tableau embed loads and is functional
- Check for double scrolling (the embed scrolls within the page) 

### Data
- Click each button
- Click a FAQ for functionality
- References table embed loads
    - Check that downloads are disabled.
- Data download link provides access to a downloadable CSV.
    - Download the CSV
- "terms of use" button

### Publications
- Cards all load with images
- Carousel is clickable
- Try 2 random items on the page for functional links

### About
- Page loads
- Scroll to bottom

### Check serotracker.com/broken
- 404 page loads and animation works

### Check serotracker.com/Canada 
- Check tableau embed loads and is functional
- Check for double scrolling (the embed scrolls within the page) 
- Check in French

PR #348, opened 2022-02-17
Windows installation details and fixes

grbdeveaux:
Added a Table of Contents.
Removed Node and NPM from the requirements table since we're documenting how to install them.
Updated the instructions for the installation of Node and NPM. I simplified the instructions, removed the instruction to check within the config file with nano. This can be tricky for new users and installation can be verified with nvm -v instead. Adding an instruction to source which should serve the same purpose.
Added a line for the installation of cURL just in case.
Included additional steps (e.g. installing other node versions) to give user a better understanding of nvm. This helped me.
Added details in the Windows section explaining the need for WSL and added links to useful guides.
Updated vscode config file.
Increased all header levels by one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,89 +1,120 @@
 
-## [Serotracker.com](https://serotracker.com/)
+# [Serotracker.com](https://serotracker.com/)
 
-### Prerequisites
 
-#### Required Software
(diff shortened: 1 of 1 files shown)


grbdeveaux:
Changes moved into PR #340 instead.

